PR #34, opened 2021-11-04
refactor standardize and center

strengejacke:


strengejacke:
This PR:

1. moves the helper functions for `center()` and `standardize()`, which were spread across the two R files, but used by both functions, into a separate R file, which now only contains the helper functions.
2. redundant code in both `center()` and `standardize()` was refactored and moved into a new helper function (see https://github.com/easystats/datawizard/pull/34/files#diff-e6cf4a64de202c987037db5a94e1be0fe1fee5bf5cc03b0c4902bacb8e11746c, function `.process_std_args()`), to avoid duplicated code.
3. Some functionality that was available in `standardize()`, but not `center()`, was added to `center()` as well.
4. Order of arguments from the data frame methods of  `center()` and `standardize()` were harmonized.

strengejacke:
(and I added some more tests)

bwiernik:
Nice! Can we store the center and scale values as attributes that we might be able to access later, such as when plotting ?

IndrajeetPatil (approved):
Awesome! 
Since all old tests pass as they are, the behavior hasn't changed, and so the refactoring looks good to me! 

Thanks for adding more tests.

strengejacke:
There's still a minor issue I think, waiting for checks. Maybe related to the new added attributes to `center()`.

PR #35, opened 2021-11-05
harmonize options for std and center

strengejacke:


strengejacke:
There's no reason why we shouldn't have the same options for `center()` as for `standardize()`, so I added `force` and `reference` arguments. What is missing is `center()` for grouped data frames. I'll work on that.

PR #47, opened 2022-01-04
more winsorizing options?

mattansb:
@IndrajeetPatil Would you be open to one or more of the following while I'm at it:

- [ ] Allow hard values for windzorizing, with `threshold = c(bottom, top)`?
- [ ] For factors, can combine levels with less than `threshold`% observations? Is that weird?


IndrajeetPatil:
Can you please create an issue about this where we can discuss this, and then you link this PR to it? That way we will have a written record in issues about what motivated this change. :)

Thanks!

strengejacke:
we could add select helpers here, too. See `ranktransform()` etc.

IndrajeetPatil:
@mattansb Do you plan to work on this any time soon?

https://github.com/easystats/datawizard/issues/133 is waiting on this PR.

mattansb:
My deadline for sending out my dissertation is 30/9, so before that I won't have time for anything other than a patch here or there...

IndrajeetPatil:
Cool. Marking this as a draft PR then. We can revisit this after #133.

mattansb:
Oh, I totally forgot about this... um, probably not...

PR #51, opened 2022-01-19
closes #50

### LINKED ISSUE #50: Implement `is_empty_object()` function
Multiple packages in our ecosystem have this same function as an internal function, and so there is unnecessary duplication of code. We can define it only once in `{datawizard}` and other packages can start using it.

```r
# is string empty?
.is_empty_object <- function(x) {
  if (is.list(x)) {
    x <- tryCatch(
      {
        .compact_list(x)
      },
      error = function(x) {
        x
      }
    )
  }
  
  if (inherits(x, c("tbl_df", "tbl"))) x <- as.data.frame(x)
  x <- suppressWarnings(x[!is.na(x)])
  length(x) == 0 || is.null(x)
}

# remove NULL elements from lists
.compact_list <- function(x, remove_na = FALSE) {
  if (remove_na) {
    x[!sapply(x, function(i) length(i) == 0 || is.null(i) || (length(i) == 1 & is.na(i)) || any(i == "NULL", na.rm = TRUE))]
  } else {
    x[!sapply(x, function(i) length(i) == 0 || is.null(i) || any(i == "NULL", na.rm = TRUE))]
  }
}
```

IndrajeetPatil:
Adds two new functions


strengejacke:
This is the insight-implementation, right? Looks good!

IndrajeetPatil:
> This is the insight-implementation, right?

Yes, the one with better performance.

bwiernik:
The ones in insight are internals and it makes more sense to export them here because data--that's the thinking right?

Looks good to me

IndrajeetPatil:
Since the insight package is now the lowest level dependency, I am wondering if it's more appropriate to export these functions from insight and then use it in other packages?

Otherwise, insight will have it own copy of these functions, and datawizard will have one, and we will need to constantly keep them in sync. What do you think?

bwiernik:
As a user, I would be much more likely to look for these functions in datawizard. I think it's okay to have internal duplicates as needed in insight. To keep them in sync, we can consider setting up a git workflow to copy them from datawizard to insight when changed?

strengejacke:
> To keep them in sync, we can consider setting up a git workflow to copy them from datawizard to insight when changed?

That sounds good to me!

IndrajeetPatil:
I am afraid to merge this now since the other two PRs I made had to be reverted. 😢 

I am going to let you guys decide if and when to merge this.

bwiernik:
Let's put all of these into one file called datawizard-internals.R

IndrajeetPatil:
> datawizard-internals.R

Why internals? Won't that be confusing since we are exporting these functions?

bwiernik:
Sorry, let me flesh out my thinking more. Let's do the following:

In datawizard:
1. Put the guts of these functions as internal functions. Name it something like shared-internals.R
2. In the current script, assign the internals to their exported names and export the functions

In insight:
1. Move these shared internals all to a file called shared-internals.R or datawizard-internals.R

IndrajeetPatil:
I have created a new issue for this: https://github.com/easystats/easystats/issues/201

Merging now since these functions are going to be included in this package irrespective of what we decide on dealing with duplication of code.

PR #52, opened 2022-01-22
add `compact_character()`

IndrajeetPatil:
@strengejacke Can you please have a look at the tests and check that the behavior and see if this is expected?

I am a bit mystified by how the function is dealing with `NULL`s and `NA`s. Something that needs to be included in the documentation.


bwiernik:
I would expect NAs to be removed too, no?

R/compact_character.R
```diff
@@ -0,0 +1,20 @@
...
+#' @examples
+#' compact_character(c("x", "y", NA))
+#' compact_character(c("x", NULL, "", "y"))
+#' compact_character(list("x", "y", NA))
+#' compact_character(list("x", NULL, "", "y"))
+#'
+#' @export
+
+compact_character <- function(x) {
+  x[!sapply(x, function(i) nchar(i) == 0 || is.null(i) || any(i == "NULL", na.rm = TRUE))]
```
bwiernik, line 18:
```suggestion
  x[!sapply(x, function(i) nchar(i) == 0 || is.null(i) || all(is.na(i)) || any(i == "NULL", na.rm = TRUE))]
```

bwiernik: Or maybe any() there?

IndrajeetPatil: > I would expect NAs to be removed too, no?

Yeah, me too, but @strengejacke has been using this internal helpers for a long time, so I am not sure if this is deliberate behavior or something that has gone unnoticed thus far.

strengejacke:
I'm not sure about the different contexts where I used this function,  but I remeber that the `"NULL"` came from:

``` r
deparse(substitute(NULL))
#> [1] "NULL"
```

<sup>Created on 2022-01-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
I also don't think that `is.null()` is doing anything.

bwiernik:
list("x", NULL)

Main question is whether NA should be removed. I think it should, right?

strengejacke:
> list("x", NULL)

That goes into `compact_list()`, not character.

bwiernik:
There are 2 examples with list() for this function

IndrajeetPatil:
I updated examples and tests. I think now the behavior is as expected.

strengejacke:
You don't need NULL in the vector, it's removed automatically. But in general: should that function also compact _lists_ of character vectors? I'd say the only valid argument is a character vector / string.

R/compact_character.R
```diff
@@ -0,0 +1,18 @@
...
+#'
+#' @param x A single character or a vector of characters.
+#'
+#' @return
+#'
+#' A character or a character vector with empty strings removed.
+#'
+#' @examples
+#' compact_character(c("x", "y", NA))
+#' compact_character(c("x", NULL, "", "y"))
```
strengejacke, line 11: no need for NULL here

R/compact_character.R
strengejacke: I think we can have `any()` here. Not sure if it will break code, but I think we can just "re-write" this from scratch.

R/compact_character.R
IndrajeetPatil: Good point. I was just copying and pasting the list examples and replacing with vectors. Modified.

bwiernik:
But NA values should be removed, no?

IndrajeetPatil:
Not sure if you are responding Daniel, or to me, but the function now removes `NA`s

https://github.com/easystats/datawizard/blob/957764e067eba714ee70a250874096b0be9b30c6/tests/testthat/test-compact-character.R#L2-L4

IndrajeetPatil:
May I merge this?

PR #54, opened 2022-01-22
WIP: closes #53

### LINKED ISSUE #53: Remove deprecated `change_scale()` function
You should be using `data_rescale()` instead. I will be making this change in the next-to-next version.

https://github.com/easystats/datawizard/blob/a766b701cf255208ff4d87685f0c77b12f353b73/R/data_rescale.R#L41-L48

cc @strengejacke, @mattansb, @DominiqueMakowski You need to make these changes in _see_, _parameters_, _effectsize_, and _modelbased_.  

https://github.com/search?p=1&q=org%3Aeasystats+change_scale&type=Code

- [x] see
- [x] parameters
- [x] modelbased
- [x] effectsize

IndrajeetPatil:
To do:
Make entry in NEWS.md about this breaking change


IndrajeetPatil:
This will remain a draft until two new `easystats` cycles are complete.

bwiernik:
What's the motivation for removing the `change_scale()` alias? I agree with using `data_rescale()` internally consistently, but having the alias available makes sense to me. I personally really like this alias and would prefer we keep it as an option (similar to `parameters()` vs `model_parameters()`).

Especially inside another function call this looks much more natural to me:

```r
mtcars |>
  dplyr::mutate(mpg_pomp = change_scale(mpg, to = c(0, 100))
```

IndrajeetPatil:
@bwiernik I created this draft PR before #57 was opened and was motivated by reducing the number of aliases. 

But, in light of the discussion in #57, this might not be worth it since we are moving towards using even more aliases to kill two birds in one stone: have a consistent naming schema and be backward compatible.

PR #61, opened 2022-02-09
adds `data_rotate` function

IndrajeetPatil:
Adapted from `{sjmisc}`


R/data_rotate.R
```diff
@@ -0,0 +1,66 @@
...
+
+
+  # add rownames as column, if requested
+  if (!is.null(rownames)) data <- rownames_as_column(data, var = rownames)
+
+  # add back attributes. therefore, delete the common attributes, like class etc.
+  # and then add attributes to our final dataframe
+
+  # a[c("names", "row.names", "class", "dim", "dimnames")] <- NULL
+  a[names(a) %in% names(attributes(data))] <- NULL
```
strengejacke, line 62: Here I would use `modifyList()`. I just didn't know of that function on the past.

R/data_rotate.R
```diff
@@ -51,16 +50,14 @@ data_rotate <- function(data, rownames = NULL, colnames = FALSE) {
...
-  # add back attributes. therefore, delete the common attributes, like class etc.
-  # and then add attributes to our final dataframe
-
-  # a[c("names", "row.names", "class", "dim", "dimnames")] <- NULL
+  # delete the common attributes
   a[names(a) %in% names(attributes(data))] <- NULL
-  attributes(data) <- c(attributes(data), a)
+
+  # and then add other attributes to the dataframe
+  attributes(data) <- modifyList(attributes(data), a)
```
strengejacke, line 60: to be more precise, it's `utils::modifyList()`, I think. ;-)

R/data_rotate.R
```diff
@@ -0,0 +1,63 @@
...
+  a <- attributes(data)
+
+  # rotate dataframe by 90 degrees
+  data <- as.data.frame(t(as.data.frame(data)))
+
+  # add column names, if requested
+  if (!is.null(colnames)) {
+    # check if we have correct length of column names
+    if (length(colnames) != ncol(data)) {
+      warning(
```
IndrajeetPatil, line 44: @strengejacke Can you think of an example where this warning might be appropriate? 
Will be good to test this condition.

strengejacke: Hm, not sure what the reason was to include this check... Maybe one column data frames?

IndrajeetPatil: Hmm, even in that context no warning is issued:

``` r
datawizard::data_rotate(dplyr::select(mtcars, wt), colnames = TRUE)
#>  [1] 2.62  2.875 2.32  3.215 3.44  3.46  3.57  3.19  3.15  3.44  3.44  4.07 
#> [13] 3.73  3.78  5.25  5.424 5.345 2.2   1.615 1.835 2.465 3.52  3.435 3.84 
#> [25] 3.845 1.935 2.14  1.513 3.17  2.77  3.57  2.78 
```

<sup>Created on 2022-02-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

strengejacke: yeah. Maybe it's only useful if we would allow a character vector for `colnames` as well.

IndrajeetPatil: Yes, definitely useful in that context. I will implement that change in a separate PR.

PR #62, opened 2022-02-12
add `replace_nan_inf()`

IndrajeetPatil:


strengejacke:
Using lapply, makes it a bit faster:

``` r
df <- data.frame(
  x = c(1, NA, 5, Inf, 2, NA),
  y = c(3, NaN, 4, -Inf, 6, 7),
  stringsAsFactors = FALSE
)


f1 <- function(data) {
  for (i in colnames(data)) {
    # convert `NaN` and `Inf` to missing
    data[[i]][is.nan(data[[i]])] <- NA
    data[[i]][is.infinite(data[[i]])] <- NA
  }
}



f2 <- function(data) {
  data[] <- lapply(data, function(i) {
    i[is.nan(i)] <- NA
    i[is.infinite(i)] <- NA
    i
  })
}


microbenchmark::microbenchmark(f1(df), f2(df), times = 1000)
#> Unit: microseconds
#>    expr  min   lq    mean median   uq    max neval cld
#>  f1(df) 66.7 70.6 92.8339   72.8 75.5 6124.8  1000   b
#>  f2(df) 38.6 41.8 52.7578   43.0 45.1 4633.2  1000  a
```

<sup>Created on 2022-02-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Issue #67, opened 2022-02-13
set_na()


strengejacke:
If we're going to copy some useful functions from sj* packages, we should also have a `set_na()` function. This is a quite common operation after importing data from Stata or SPSS.

IndrajeetPatil:
Is this the same as `replace_na()`?

strengejacke:
No, I think it's the counter-part.

replace_na = replace NA with a value, i.e. turn NAs into a non-NA value
set_na = replace values with NA, i.e. replace specific values with NA.

IndrajeetPatil:
Ah, I see. Hadn't come across it.

How do you feel about naming it `convert_to_na()`?

PR #68, opened 2022-02-13
convert_to_na

strengejacke:
#67


strengejacke:
should be finished now, just missing tests.

IndrajeetPatil:
Can you please also add an entry about this change in `NEWS.md`? Feel free to directly push to `master`.

PR #74, opened 2022-02-14
Recode fun

strengejacke:


strengejacke:
Name of that function is just a suggestion, we can have any other name, if you like. Would also like to add a data frame and possibly factor method.

strengejacke:
@IndrajeetPatil I think this is quite complete now (as initial draft). Proper docs and tests are missing.

IndrajeetPatil:
Thanks, will have a closer look later during the day.

strengejacke:
@easystats/maintainers (does this work?)

@mattansb @bwiernik @DominiqueMakowski @vincentarelbundock What do you say about having a recode-function? Any suggestions for the name of this function? It does not do classical "value"-replacement, but rather stuff like dichotomizing / grouping. Thus, `data_recode()` is probably not exact / clear enough here. 

Here are some examples:

``` r
library(datawizard)
set.seed(123)
x <- sample(1:10, size = 500, replace = TRUE)

table(x)
#> x
#>  1  2  3  4  5  6  7  8  9 10 
#> 45 42 47 46 44 52 56 52 57 59

# by default, at median
table(data_recode(x))
#> 
#>   1   2 
#> 276 224

# into 3 groups, based on distribution (quantiles)
table(data_recode(x, split = "quantile", n_groups = 3))
#> 
#>   1   2   3 
#> 180  96 224

# into 3 groups, try to return similar group sizes
# (i.e. similar count for each value/level)
table(data_recode(x, split = "equal_size", n_groups = 3))
#> 
#>   1   2   3 
#> 134 198 168

# into 3 groups, manual cut offs
table(data_recode(x, split = c(3, 5)))
#> 
#>   1   2   3 
#> 134  90 276

set.seed(123)
x <- sample(1:100, size = 500, replace = TRUE)

# into 5 groups, try to return similar group sizes
table(data_recode(x, split = "equal_size", n_groups = 5))
#> 
#>   1   2   3   4   5 
#>  98 101  97  99 105

# into 5 groups, try to return same range within groups
# i.e. 1-20, 21-40, 41-60, etc.
table(data_recode(x, split = "equal_range", size_groups = 20))
#> 
#>   1   2   3   4   5 
#>  89 116  96  94 105


# data frames
head(data_recode(iris))
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1            1           2            1           1  setosa
#> 2            1           1            1           1  setosa
#> 3            1           2            1           1  setosa
#> 4            1           2            1           1  setosa
#> 5            1           2            1           1  setosa
#> 6            1           2            1           1  setosa

head(data_recode(iris, append = TRUE))
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species Sepal.Length_r
#> 1          5.1         3.5          1.4         0.2  setosa              1
#> 2          4.9         3.0          1.4         0.2  setosa              1
#> 3          4.7         3.2          1.3         0.2  setosa              1
#> 4          4.6         3.1          1.5         0.2  setosa              1
#> 5          5.0         3.6          1.4         0.2  setosa              1
#> 6          5.4         3.9          1.7         0.4  setosa              1
#>   Sepal.Width_r Petal.Length_r Petal.Width_r
#> 1             2              1             1
#> 2             1              1             1
#> 3             2              1             1
#> 4             2              1             1
#> 5             2              1             1
#> 6             2              1             1

head(data_recode(iris[3:5], force = TRUE))
#>   Petal.Length Petal.Width Species
#> 1            1           1       1
#> 2            1           1       1
#> 3            1           1       1
#> 4            1           1       1
#> 5            1           1       1
#> 6            1           1       1
```

<sup>Created on 2022-02-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
The function seems fine. For a name, I suggest `data_cut()`

How does its functionality differ from the base R `cut()`? https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/cut

strengejacke:
> The function seems fine. For a name, I suggest `data_cut()`
> 
> How does its functionality differ from the base R `cut()`? https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/cut

It all boils down to using `cut()`, but it simplifies (or automates) the definition of cut-points / breaks, thus making this more accessible to users not coming from R.

mattansb:
Seems useful! 
I would change the name of the function though - "recode" feels like it's something else...

bwiernik:
Yeah, I wouldn't expect that this is what a function called "recode" would do

vincentarelbundock:
This looks super cool. Well done!

I don't have thoughts on this specific function design which, as I said, is very cool. However, I do have to say that I am starting to worry a bit about sprawl in the `easystats` ecosystem. At some point in the future, someone is going to have less energy for this, and maintaining a billion functions could become a problem for long term sustainability. Another potential issue is that using a lot of `datawizard` abstractions throughout the `easystats` codebase may make it harder for newcomers (like me) to wrap their heads around the meaning of the code (especially when a ~5 lines of Base R would do).

Again, I don't mean to be negative. I'm just raising something I've been worried about for a bit...

strengejacke:
> This looks super cool. Well done!
> 
> I don't have thoughts on this specific function design which, as I said, is very cool. However, I do have to say that I am starting to worry a bit about sprawl in the `easystats` ecosystem. At some point in the future, someone is going to have less energy for this, and maintaining a billion functions could become a problem for long term sustainability. Another potential issue is that using a lot of `datawizard` abstractions throughout the `easystats` codebase may make it harder for newcomers (like me) to wrap their heads around the meaning of the code (especially when a ~5 lines of Base R would do).
> 
> Again, I don't mean to be negative. I'm just raising something I've been worried about for a bit...

I share your concern to some extent, however, it's sometimes not just

> (especially when a ~5 lines of Base R would do).

`data_recode()` (or whatever name it will have) is imho more accessible to newcomers, and comes with a nice data frame methods as well. And it preserves variable labels (which is a very important missing feature in base R's functions). :-)

Same with `data_merge()`. `merge(sort = FALSE)` not always works as expected, `all.x` and `all.y` and `by.x` and `by.y` is confusing to newcomers, and `data_merge()` offers some additional join-methods that you don't have with base R's `merge()` (and `data_merge()` also preserves variable and value labels...).

I think these functions can be beneficial to newcomers, they work in "pipe" workflows, and since they are wrapped around base R functions, maintenance should be ok once we have a stable base.

> codebase may make it harder for newcomers

Yeah, that might be true. But balancing accessibility of R for users vs. accessibility of easystats-codebase for contributors, I would probably give the first one a bit more weight :-)

strengejacke:
what about `data_group()` / `data_grouping()` and `data_cut()` as alias?

strengejacke:
(actually, I came to the idea of implementing `data_merge()` and `data_cut()`/`data_recode()` because I'm currently working with [share-data](http://www.share-project.org/) again, and am just doing the data preparation stuff. I used to use `merge_df()` and `split_var()` etc. from _sjmisc_, but now since we have _datawizard_, I thought it might be nice to have a "feature complete" package in this regard as well)

bwiernik:
`data_bin()` or `data_cut()`

Group evokes `dplyr::group_by()`

PR #76, opened 2022-02-20
Add data_reverse() and document how to reverse-score using data_rescale()

bwiernik:


R/data_reverse.R
```diff
@@ -0,0 +1,197 @@
...
+  if (is.null(select)) {
+    select <- names(x)
+  }
+
+  if (!is.null(exclude)) {
+    select <- setdiff(select, exclude)
+  }
+
+  # Transform the range so that it is a list now
+  if (!is.null(range)) {
```
IndrajeetPatil, line 187: I think you can avoid this nesting by using short-circuiting `&&`:

```r
if (!is.null(range) && !is.list(range)) {...}
```

R/data_reverse.R
```diff
@@ -0,0 +1,197 @@
...
+
+  # check for formula notation, convert to character vector
+  if (inherits(select, "formula")) {
+    select <- all.vars(select)
+  }
+  if (inherits(exclude, "formula")) {
+    exclude <- all.vars(exclude)
+  }
+
+  # dplyr < 0.8.0?
```
IndrajeetPatil, line 138: 0.8.0 was released in Feb 2019 and is 3 years old already.

Should we continue to support this version? I don't think so. I think we can safely assume that most, if not all, of our users are using a version `> 1.0.0`.

@DominiqueMakowski This comment also applies for a lot of conditionals we have in `{report}` package.

IndrajeetPatil (approved):
Thanks, @bwiernik! 

I will let Dom and Daniel have a closer look at this, but I had a few minor points.

Additionally, can you please:
- add a few tests
- make an entry in `NEWS.md`

R/data_reverse.R
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  # Transform the range so that it is a list now
+  if (!is.null(range)) {
+    if (!is.list(range)) {
+      range <- stats::setNames(rep(list(range), length(select)), select)
+    }
+  }
+
+  x[select] <- as.data.frame(sapply(select, function(n) {
```
strengejacke, line 193: Can't we just use `lapply()` instead of `as.data.frame(sapply())` here?

R/data_reverse.R
strengejacke: Yes, it might be safe to remove that conditional.

IndrajeetPatil:
@bwiernik Do you think this will be ready sometime soon? 

I would like to make `0.3.0` release soon since there is a lot of new functionality that would be good to have in the CRAN version ASAP.

bwiernik:
If you could write a test, it can be ready immediately, otherwise I can try to do it this weekend

IndrajeetPatil:
We can add tests later.

PR #80, opened 2022-03-03
Implement and export rownames/id to column and inverse

### LINKED ISSUE #2: function: rownames/id to column and inverse
ala `tibble::rowids_to_column()` and `tibble::rownames_to_column()`

etiennebacher:
This PR is related to #2. It exports the function `rownames_as_column()`, and implements and exports `column_as_rownames()`. I added sanity checks and tests, but let me know if something is missing or should be modified. I also added a NEWS entry.

Close #2.


strengejacke:
Thanks, looks good! Two things from my side:

1. I think `Argument 'var' must be a character.` is not the correct wording, I think it should probably be `string` (a character is just one char).
2. I think stopping is too radical, probably we should yield a warning and return the original input.

etiennebacher:
Thanks,

1. Fixed
2. I'm not sure, `tibble::column_to_rownames()` stops with an error message:
```r
library(tibble)

mtcars |> 
  rownames_to_column() |> 
  column_to_rownames(var = 1)

#> Error:
#> ! Can't find column `1` in `.data`.
#> Run `rlang::last_error()` to see where the error occurred.
```

bwiernik:
Re: wording, We use "of type character" throughout our docs, rather than "string", as does tidyverse, so let's go with that

bwiernik:
It would make sense to me if column_as_rownames() permitted a numeric index as well as a character value

strengejacke:
> It would make sense to me if column_as_rownames() permitted a numeric index as well as a character value

Yes, agreed.

etiennebacher:
Done for both points, not sure about the new error messages in `column_as_rownames()` though

strengejacke:
Thanks, looks good! If noone else has any objections against stopping instead of just warning, this can be merged.

IndrajeetPatil:
Should we not be trying to avoid any namespace collision with _tidyverse_ packages?

The `rownames_as_column()` function will now be present both in _tibble_ and _datawizard_ and there will be masking issues. I think we need to choose a different name for this function.

R/utils_data.R
```diff
@@ -7,10 +7,60 @@
   x
 }
 
+#' Tools for working with row names
+#'
+#' @param x A data frame.
+#' @param var Name of column to use for rownames. For \code{column_as_rownames()}, this argument can be the variable name or the column number.
+#'
```
IndrajeetPatil, line 14: We try to use markdown syntax as much as possible in roxygen documentation.

So, for example, here, instead of `\code{column_as_rownames()}`, you can use ``column_as_rownames()``.

etiennebacher:
> Should we not be trying to avoid any namespace collision with _tidyverse_ packages?
> 
> The `rownames_as_column()` function will now be present both in _tibble_ and _datawizard_ and there will be masking issues. I think we need to choose a different name for this function.

`tibble` has `column_to_rownames()` and `rownames_to_column()`, whereas here it is `column_as_rownames()` and `rownames_as_column()` so no problem of namespace collision.

tests/testthat/test-utils_data.R
```diff
@@ -0,0 +1,50 @@
+test_that("rownames_as_column works", {
```
IndrajeetPatil, line 1: Great job with the tests! Thanks.

tests/testthat/test-utils_data.R
```diff
@@ -0,0 +1,50 @@
...
+  test <- rownames_as_column(mtcars, "new_column")
+  expect_true("new_column" %in% names(test))
+  expect_true(test[1, "new_column"] == "Mazda RX4")
+})
+
+test_that("rownames_as_column doesn't work if var is not a character", {
+  expect_error(rownames_as_column(mtcars, var = 1))
```
IndrajeetPatil, line 8: If you want, you can also add an expected error message to `regexp` argument in `expect_error()`, so that we know that the functions are not working for expected reasons.

IndrajeetPatil (changes requested):

IndrajeetPatil:
> so no problem of namespace collision

@etiennebacher Glad to hear that!

IndrajeetPatil:
@etiennebacher If you want, you can also fix the pkgdown build issues with an update `_pkgdown.yml` to include these new functions.

You can include the following code

```yml
  - title: Rowname helpers
  - contents:
      - contains("rownames")
```

below this part

https://github.com/easystats/datawizard/blob/02e30d627f38a929c9d9f61067aa72877ad6d02b/_pkgdown.yml#L62-L65

NEWS.md
```diff
@@ -6,6 +6,10 @@ The following functions are now re-exported from `{insight}` package:
 `object_has_names()`, `object_has_rownames()`, `is_empty_object()`,
 `compact_list()`, `compact_character()`
 
+NEW FUNCTIONS
+
+  * To convert rownames to a column, and vice-versa: `rownames_as_column()` and `column_as_rownames()`.
```
IndrajeetPatil, line 11: In the parentheses, please also mention your GitHub handle and PR number, so `(@etiennebacher, #80)`.

tests/testthat/test-utils_data.R
etiennebacher: Good idea! I didn't know there was a regexp argument

NEWS.md
etiennebacher: should I also mention issue #2 ?

IndrajeetPatil: Not needed. The PR already contains that information.

IndrajeetPatil (approved):
Thanks for the great PR, @etiennebacher!

IndrajeetPatil:
I will merge this once the checks are completed.

etiennebacher:
Thanks for the great packages and the feedback for this PR! Maybe see you on future ones ;)

PR #96, opened 2022-03-08
remove useless check in column_as_rownames()

### LINKED ISSUE #95: small change to make in column_as_rownames()
In `column_as_rownames()`, the behavior implemented in #80 is to use "rowname" as default if `var = NULL`. This doesn't work because:
https://github.com/easystats/datawizard/blob/817d90aa3bba501435cfceb55c6309ecc9b190f5/R/utils_data.R#L48-L50
comes after the checks with `is.character()` and `is.numeric()` which means that this condition is never called if `var = NULL`. All that is needed is to put the `NULL` check first, but I'm not sure whether using "rowname" if `var = NULL` should be the default behavior. 

Basically there is something to change but it depends on your decision:
* either keep this behavior but put the `NULL` check first (+ add a test for when `var = NULL`)
* remove this behavior and therefore also this condition because it is useless as is

I'm ok to do a PR for this if you want

etiennebacher:
Remove condition when `var = NULL` in `column_as_rownames()` as this case is already taken into account by sanity checks.

Close #95

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -12,7 +12,8 @@ Authors@R: c(
     person("Mattan S.", "Ben-Shachar", , "matanshm@post.bgu.ac.il", role = "aut",
            comment = c(ORCID = "0000-0002-4287-4801")),
     person("Brenton M.", "Wiernik", , "brenton@wiernik.org", role = "aut",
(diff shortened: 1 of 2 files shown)


IndrajeetPatil (approved):
Thanks!

PR #109, opened 2022-03-15
Add tests for data_reverse()

etiennebacher:
I don't know if this should be in NEWS since it has no influence on user experience, let me know


IndrajeetPatil:
> I don't know if this should be in NEWS since it has no influence on user experience, let me know

You are right, this shouldn't be logged in `NEWS`.

strengejacke (approved):

IndrajeetPatil:
@etiennebacher Thanks a ton!!

PR #112, opened 2022-03-17
Convert to na select helpers

strengejacke:


strengejacke:
@IndrajeetPatil I realized two things, first, we might have an overview of functions that could use select helpers, too. and 2), we probably need to take care of the description of the arguments. Select-helpers are handled differently in `data_extract()` and the other functions (which is ok), and not all functions have them documented (like reshape).

IndrajeetPatil (approved):

PR #114, opened 2022-03-18
init

strengejacke:


strengejacke:
select helpers for `data_addprefix()` and suffix

R/convert_to_na.R
```diff
@@ -125,7 +125,7 @@ convert_to_na.data.frame <- function(x, na = NULL, select = NULL, exclude = NULL
   .attach_packages(conflicting_packages)
 
   # return valid column names, based on pattern
-  select <- .evaluated_pattern_to_colnames(select, data, ignore_case, verbose, exclude)
+  select <- .evaluated_pattern_to_colnames(select, x, ignore_case, verbose, exclude)
```
strengejacke, line 128: @IndrajeetPatil  @DominiqueMakowski Sometimes, we use `data`, sometimes `x`. Can we decide on a unique name for the data argument?

IndrajeetPatil: If the function works primarily with dataframes, then `data`; if it works with vectors, then `x`.

If both, then it's tricky. But I guess `data` should be fine. It's generic enough to accommodate different data structures.

strengejacke:
Feel free to merge. This PR adds more select-helper-functionality to some functions, improves the documentations, and also allows formulas as `select` (or `pattern`). Some tests have also been added.

IndrajeetPatil (approved):
Thanks, Daniel!

PR #117, opened 2022-03-18
Data cut grouped df

strengejacke:


strengejacke:
#118 needs to be merged first.

PR #119, opened 2022-03-18
Use refactored NSE function

strengejacke:


R/data_remove.R
```diff
@@ -10,10 +10,10 @@ data_remove <- function(data, pattern, ignore_case = FALSE, verbose = FALSE, ...
   ## TODO set verbose = TRUE by default in a later update?
 
   # evaluate arguments
-  pattern <- .select_nse(pattern, data, exclude = NULL, ignore_case)
```
strengejacke: @mattansb when the variable is named `pattern` instead of `select`, evaluation doesn't work... Any ideas?

strengejacke:
Ok, for future reference: Any pattern of `select` (i.e. literale name, character vectors, numeric, select-helpers, `:` ranges...) will be evaluated in `.select_nse(select, data, exclude = NULL, ignore_case)`, and the return value are *all* matching, valid, full column names in `data`, or an empty vector for no match. This makes column matching pretty easy for the functions.

Only limitation seems to be (see https://github.com/easystats/datawizard/pull/119#discussion_r830374448) that the argument name from the function needs to be `select`. When the argument has a different name, full evaluation inside `.select_nse()` fails.

mattansb:
> @mattansb when the variable is named pattern instead of select, evaluation doesn't work... Any ideas?

NSE is hard......

PR #122, opened 2022-03-19
.select_nse: support arg exclude and add tests

etiennebacher:


etiennebacher:
Got mixed up with convert_na_to branch, will open a new one for select_nse only

Issue #120, opened 2022-03-19
Revise API / arguments


strengejacke:
1. Is there a particular reason why we need to extract `row.names` in `data_extract()`? If not, we could just simplify that function and the the select-helpers-design from https://github.com/easystats/datawizard/pull/119.

2. Do we need to have the `cols` argument to be named `cols`? (like in `data_to_long()`) Else, we could simplify the code, using select-helpers there, too, and harmonizing / simplifying the _datawizard_ API in general.

3. Same for `data_relocate()` and `data_reorder()`: we could simply rename `cols` into `select` and simplify the code base.

DominiqueMakowski:
imo we can rename cols to select everywhere yeah

IndrajeetPatil:
Even if this causes a breaking change, I think it's worth it if we get a more consistent API out of it.

I will assess the damage in the ecosystem and make patches where needed.

strengejacke:
I'm also having teaching easystats in mind, and the fewer different names we need for arguments, the easier it is to learn. Even if the underlying functionality doesn't change, it is less confusing when the help file doesn't mention too many arguments

strengejacke:
@IndrajeetPatil Where can I see which code parts are not yet covered from tests?

IndrajeetPatil:
Here: https://app.codecov.io/gh/easystats/datawizard

strengejacke:
I think most critical will be `reshape_longer()`: https://github.com/search?q=org%3Aeasystats+reshape_longer&type=code

But maybe we can keep `cols` as an alias for some time, so it won't break code?

IndrajeetPatil:
> But maybe we can keep cols as an alias for some time, so it won't break code?

Yes, I think that's a good solution. 

Before next releases of other packages, I will make sure to replace all `cols = ` instances with `select = `.

strengejacke:
All three points addressed. We could probably rename the first argument `data` to `x` for some functions, to have more consistency across functions.

PR #124, opened 2022-03-19
reshape_longer select helpers

strengejacke:
#120


strengejacke:
I think we haven't used the "select-helper" functionality for `cols` in other packages yet (wasn't implemented before), so this should not be a breaking change. We may check by setting `Remotes` to this PR in other packages?

IndrajeetPatil:
https://github.com/search?q=org%3Aeasystats+reshape_longer&type=Code

We are the only ones using this function, so we can fix if we find any breakages. 

But the changes so far look backward-compatible, so no need to worry.

I will run a reverse dependency check just to be sure once you merge this.

PR #126, opened 2022-03-20
data_match: return filtered data by default

DominiqueMakowski:
@strengejacke I think you're right, returning the filtered should be default behavior as it's more consistent with the other `data_*` functions. I'll open PRs in the repos (mostly modelbased) where it needs to be fixed according to this (breaking) change, and then we can merge this

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 MAJOR CHANGES
 
+* `data_match()` now returns filtered data by default. Old behavior (returning rows indices) can be set by setting `return_indices = TRUE`.
+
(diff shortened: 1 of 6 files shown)


DominiqueMakowski:
In fact that's it there was only once usage of this function in modelbased. In insight there is a local (duplicate) `.data_match()` helper in `get_datagrid`, but it's there so we don't require datawizard

IndrajeetPatil (approved):
Yeah, I agree that we should be defaulting to returning a dataframe.

Thanks for making this change!

Issue #125, opened 2022-03-20
Export select helper


DominiqueMakowski:
The current `.select_nse` is not exported, yet it's likely that we will need it throughout the verse. Should we export it directly? Or somehow allow its usage via `data_findcols()`, which was initially meant to fulfil that role:

https://github.com/easystats/datawizard/blob/da3e9f8769f245c34231f2e62c7eccf91fcbeb22/R/data_findcols.R#L22-L31

strengejacke:
Yes, this is what I thought, too, when I added select helpers to all the different functions. When we would change `data_findcols()`, it ends up just calling the select-helper function. I think we should change data_findcols, too.

strengejacke:
This will break code in some places:

https://github.com/search?q=org%3Aeasystats+data_findcols&type=code

We can do two things: add arguments `select` and `exclude` (and the end of the argument list) and keep the old arguments for now, so we have a smooth transition. Or, simply add a new function and deprecate `data_findcols()` later?

DominiqueMakowski:
> add arguments select and exclude 

I am okay with just replacing and having a hard transition ^^, or as you say we keep the old args and deprecate them

> simply add a new function

If you have a better name in mind, sure


It seems like it's gonna be a big round of CRAN-update anyway

IndrajeetPatil:
Yeah, adding a new function might be the way to go here.

I like: `find_columns()`. 

Not `data_find_columns()`; `data_` prefix is redundant here because by definition columns can only be present in columnar data frame structure.

strengejacke:
> Not `data_find_columns()`; `data_` prefix is redundant here because by definition columns can only be present in columnar data frame structure.

Makes sense to me.

DominiqueMakowski:
I was about to say "but what about the data_* prefix consistency" but then we also have lots of `find_*` (mostly in insight) so it's good :)

PR #127, opened 2022-03-20
data_findcols gains select helpers

strengejacke:
#125


strengejacke:
This is what I would suggest:

```r
data_findcols(
  data,
  pattern = NULL,
  starts_with = NULL,
  ends_with = NULL,
  select = NULL,
  exclude = NULL,
  ignore_case = FALSE,
  verbose = TRUE,
  ...
)
```

It doesn't break any code, the help-file indicates that `pattern`, `starts_with` and `end_width` are deprecated, and for now, we must explicitly name the `select` argument, see examples:

```r
# Find columns names by pattern
data_findcols(iris, select = starts_with("Sepal"))
data_findcols(iris, select = ends_with("Width"))
data_findcols(iris, select = regex("\\."))
data_findcols(iris, select = c("Petal.Width", "Sepal.Length"))
```

Until we have passed one or two release cycles, where we then remove the arguments `pattern`, `starts_with` and `end_width`  and have `select` as second argument, like in other functions.

strengejacke:
ok. `data_findcols()` did not change, but is not "visible" anymore (but exported), and `find_columns()` is the new function that simply calls the select helper function.

strengejacke:
Looks like many chaotic changes, but it's less harmful than it appears to be ;-)
It adds `find_columns()`, restructures the docs, and adds more select-helper-functionality to data-frame methods (like `center.data.frame()` or `center.grouped_df()` etc.).

strengejacke:
Maybe we end up having all functions (where applicable) following a similar design:

`foo(data/x, select, exclude, ignore_case, ... <some other args>)`, where `select` and `exclude` can use select-helpers.

PR #140, opened 2022-03-25
improve docs for visualization_recipe

DominiqueMakowski:


R/visualisation_recipe.R
```diff
@@ -2,9 +2,17 @@
 #'
 #' This function prepares objects for visualisation by returning a list of
 #' layers with data and geoms that can be easily plotted using for instance
-#' `ggplot2`. See the documentation for your object's class:
-#'
+#' `ggplot2`.
+#' \cr\cr
```
IndrajeetPatil, line 6: No need to use ` \cr\cr` any more since we use markdown syntax to write `.Rd` docs.

IndrajeetPatil (changes requested):

R/visualisation_recipe.R
DominiqueMakowski: Without that it puts the text in the Details section

IndrajeetPatil: Maybe because there is no `#' @description` at the top, and so `roxygen2` is using the heuristic of separate sections per paragraph?

IndrajeetPatil (approved):

PR #153, opened 2022-04-13
Update _pkgdown.yml

strengejacke:


strengejacke:
@IndrajeetPatil Not sure if we can see the website from this PR, but what do you think about this structure (or in general, revising the pkgdown structure)

IndrajeetPatil (approved):
Love it! I was actually going that we restructure it, so happy to see that you've already taken care of that!

PR #158, opened 2022-04-29
#146 move standardize default to datawizard

### LINKED ISSUE #146: standardize.default()
@mattansb @bwiernik @IndrajeetPatil 

Two points:

1. The default method for `standardize()` is currently located in _effectsize_, however, it passes *all* inputs to `.standardize_models()`. Can we add a check if the input is a model object, and if not, give a warning and return the original input?

``` r
effectsize::standardize(as.complex(1:5))
#> Warning: Could not access model information.
#> Warning: Could not get model data.
#> Error in if (m_info$is_bayesian) {: argument is of length zero
```

2. When _effectsize_ is not loaded, but only _datawizard_, there is no default-method for standardize. This is problematic when applied to a data frame and this data frame contains variables of not-supported classes (like complex) - `standardize()` then fails with error. I'd prefer a message/warning and return the original input instead if it cannot be standardized. Else, we might get errors that are confusing for user. We can't add the default-method to _datawizard_ at the moment, due to a namespace conflict (overwriting). That's why I removed it again from datawizard for that reason:

```r
standardize.default <- function(x, verbose = TRUE, ...) {
  if (isTRUE(verbose)) {
    message(insight::format_message(sprintf("Standardizing currently not possible for variables of class '%s'.", class(x)[1])))
  }
  x
}
```

strengejacke:
Close #146


strengejacke:
@mattansb @bwiernik @IndrajeetPatil @DominiqueMakowski 

What do you think? _effectsize_ could re-import it from _datawizard_

mattansb (approved):
I'm assuming no changes were made - looks like the functions and tests were copied here.

Just need a version bump on datawizard.

mattansb:
Yes, effectsize will re-export. Should I add a startup message about this change?

strengejacke:
Yes, I just copied the files, only addition is the check for model objects at the beginning of `standardize.default`.

I'm not sure a startup message is required, when you re-export, because nothing changes from a user perspective, or?

bwiernik:
Nothing changes. We should add a conditional export for R>=3.6 like we do for pp_check() to avoid masking if both packages are attached

strengejacke:
> Is this going to cause masking messages if both packages are attached

Only as long as effectsize is not updated. This should just be temporary.

strengejacke:
Can we merge this? Then we can check _parameters_ and _effectsize_ on the master of _datawizard_.

IndrajeetPatil:
Should we mention this change in `NEWS.md`?

mattansb:
Yes.

PR #161, opened 2022-05-01
export `distribution_mode()`

IndrajeetPatil:


strengejacke:
Do we want to have a new function, or to integrate this in `describe_distribution.factor()`?

strengejacke:
Ah, got it. This was meant as addition to `mean` etc. Why not `mode`?

strengejacke:
What should `distribution_mode(iris$Species)` return? just one, or all three levels? @mattansb @bwiernik ?

strengejacke:
``` r
library(datawizard)
distribution_mode(iris$Species)
#> [1] setosa
#> Levels: setosa versicolor virginica

uniqv <- unique(iris$Species)
tab <- tabulate(match(iris$Species, uniqv))
idx <- which(tab == max(tab))
uniqv[idx]
#> [1] setosa     versicolor virginica 
#> Levels: setosa versicolor virginica
```

<sup>Created on 2022-05-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

bwiernik:
All three levels I think

IndrajeetPatil:
If there is a tie, the first element should be returned, right?

E.g., currently, ` distribution_mode(c(1, 2, 3, 3, 4, 4, 5))` returns `3` and not `c(3, 4)`, and the same holds for factors as well?

bwiernik:
Do we export the function for MAP? Might be goods to link to it with seeAlso and say that it may be better for continuous variables

bwiernik:
@IndrajeetPatil Yes, but I interpreted Daniel's questions about being about the factor levels. We shouldn't re-level the factor but instead preserve the original levels. 

Should we have a warning or message when there are multiple modes?

strengejacke:
Yeah, I was actually meaning how we should deal with ties, i.e. just the first, or more values. For MAP, we have `bayestestR::map_estimate()` or `bayestestR::point_estimate(centrality = "MAP")`.

bwiernik:
I think maybe the first and give a message/warning about multiple modes. Maybe with an attribute with all of them? Or maybe we make a new value like 'Setosa, Versicolor' and save the individual values as an attribute

strengejacke:
Showing first, give a warning and all as attribute sounds good.

strengejacke (approved):

elinw:
> Showing first, give a warning and all as attribute sounds good.

Was this  implemented?

strengejacke:
Yes: https://easystats.github.io/datawizard/reference/distribution_mode.html

strengejacke:
Ah, you mean the warning? No, not yet implemented.

PR #162, opened 2022-05-01
argument_order

strengejacke:
When working with _datawizard_, I realized that for data frame methods, it's probably better to have the arguments `select` and `exclude` first, followed by the function specific arguments. So all data frame methods follow the same design: 

`foo <- function(data frame, select, exclude, specific)`

What do you think?


strengejacke:
@mattansb @IndrajeetPatil @bwiernik @DominiqueMakowski What do you think? First, I thought it might be good to have the data argument, the function specific arguments, and then select/exclude, as described in our readme from _datawizard_

> The design of the {datawizard} functions follows a design principle that makes it easy for user to understand and remember how functions work:
> 
> the first argument is the data
> the following arguments are main arguments, related to the specific tasks of the functions
> further arguments can be select-helpers, which are offered for convenience reasons (so there is no need for interim calls to get_columns())

But I'm working on a paper and doing a lot of data wrangling now, and then I realized that it feels more natural to have the order 1. data, 2. select/exclude, 3. function specific args. I think this is also how tidyverse functions do it.

This PR changes the order of arguments for all grouped_df/data frame methods, and fixed a few affected tests.

IndrajeetPatil:
Love it! I also find this order to be much more natural than having function specific parameter following `x`.

strengejacke:
And I don't think this change will break any packages (not 100% sure, though)

strengejacke:
And while I'm, working, (s)he sits in front of the window-door and is waiting for nuts...

PXL_20220502_071224781 PORTRAIT

mattansb:
Agree!

bwiernik:
Two wrinkles:

1. This means the argument order for vector and data frame methods are different

2. What is the user doesn't want to apply select/exclude? Then they must name every function-specific argument

Are those concerns?

strengejacke:
> This means the argument order for vector and data frame methods are different

Yes, at least data frame methods have select/exclude inserted before the remaining common arguments.

> What is the user doesn't want to apply select/exclude? Then they must name every function-specific argument

Yes, but I wonder for how many functions this would apply? `select` and `exclude` are only used for data frame methods. Most functions are usually only useful when applied to a certain selection of variables (like `data_recode()`, `data_to_long()`, `data_cut()`, ...). For real world data, you usually don't want to apply the function to _every_ variable in the data.

However, there may be exceptions, and then it's slightly more effort to explicitly name the arguments, I'm not sure what the pro's and con's would be. If we have good defaults, then it doesn't matter when we have data-select-other order.

That said, we _could_ only have `select` after data, and then move `exclude` to the end, together with `ignore_case`, `verbose` etc,

bwiernik:
We definitely don't want to drop the argument. That would necessitate always using dplyr::across(), which we don't want. 

I don't have a big problem putting select and exclude first, just wanted to raise a potential concern

strengejacke:
I'm not 100% convinced if this is the better choice, but the last days when I used _datawizard_ "in real life" situations, my impression was it has more advantages than disadvantages. Either way, we should decide which way to go for the next update, and then probably collect some user feedback and revert the changes later, if necessary. What do you think?

bwiernik:
Most of the time when I use datawizard functions, I use the vector versions inside mutate(). I lean toward what you suggest about select/exclude here. We can always solicit feedback after

PR #166, opened 2022-05-13
CRAN release candidate 0.4.1

IndrajeetPatil:
cc @strengejacke 

Do you want to make any other changes or can I submit to CRAN?


strengejacke:
You can submit. I think the GitHub Actions have some technical issues at the moment, so check errors are probably temporary.

IndrajeetPatil:
@strengejacke Can you please check if the test failure is expected here?

IndrajeetPatil:
These failures seem legit:

```r
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure (test-data_rename.R:44:3): data_rename works when too many names in 'replacement' ──
`x <- data_rename(test, replacement = paste0("foo", 1:6))` did not throw the expected message.
── Failure (test-data_rename.R:54:3): data_rename works when not enough names in 'replacement' ──
`x <- data_rename(test, replacement = paste0("foo", 1:2))` did not throw the expected message.

[ FAIL 2 | WARN 23 | SKIP 0 | PASS 1015 ]
```

strengejacke:
I changed the default width in `insight::format_message()`, thus the line breaks have changed, I guessed. We shouldn't match to the complete message string, just a keyword.

IndrajeetPatil:
I tried twice now to submit the package, but it doesn't pass the incoming checks because the _insight_ update is not available on CRAN's Windows machines.

```r
package datawizard_0.4.1.tar.gz does not pass the incoming checks automatically, please see the following pre-tests:
Windows: <https://win-builder.r-project.org/incoming_pretest/datawizard_0.4.1_20220513_224908/Windows/00check.log>
Status: 1 ERROR
Debian: <https://win-builder.r-project.org/incoming_pretest/datawizard_0.4.1_20220513_224908/Debian/00check.log>
Status: OK
```

I also informed Uwe that this is a false positive. This is all I can do at the moment. 

This is just the first step; there is still reverse dependency check, which might also take some time.

Fingers crossed that the package isn't archived.

strengejacke:
_insight_ has been built for Windows at the weekend.

IndrajeetPatil:
Yeah. 

Now the _datawizard_ submission is stuck at manual inspection stage where someone will check if the additional M1mac issues have been resolved.

IndrajeetPatil:
And after that there will be reverse dependency check and another wait 🥵

PR #168, opened 2022-05-19
remove data_findcols()

strengejacke:


IndrajeetPatil:
@bwiernik We need to update _correlation_ on CRAN in this release because it is using `data_findcols()`, and now that the _datawizard_ update is on CRAN, users will be getting warnings about its usage. 

cf. https://github.com/easystats/correlation/issues/235

IndrajeetPatil (approved):

Issue #49, opened 2022-01-05
Allow more thresholds for windzorizing


mattansb:
*Originally posted https://github.com/easystats/datawizard/pull/47*

***Current behavior***: Thresholds are relative, based on top/bottom percentile.

***Suggested behavior***: This can be expanded in many ways:

- Fixed values - used passed lower and upper thresholds.
- Relative Z score
- Relative robust Z score

PR #179, opened 2022-06-25
Winsorize based on the MAD

rempsyc:
First attempt(!) at adding the possibility to winsorize based on the MAD, related to #177 & #49 & #47.

### DIFF
--- a/R/winsorize.R
+++ b/R/winsorize.R
@@ -19,12 +19,14 @@
 #' @param data Dataframe or vector.
 #' @param threshold The amount of winsorization.
 #' @param verbose Toggle warnings.
-#' @param robust Logical, if TRUE, winsorizing is done via the median absolute deviation (MAD).
+#' @param method One of "percentile" or "zscore".
+#' @param robust Logical, if TRUE, winsorizing through the "zscore" method is done via the median and the median absolute deviation (MAD); if FALSE, via the mean and the standard deviation.
 #' @param ... Currently not used.
 #'
 #' @examples
 #' winsorize(iris$Sepal.Length, threshold = 0.2)
-#' winsorize(iris$Sepal.Length, threshold = 3, robust = TRUE)
+#' winsorize(iris$Sepal.Length, threshold = 2, method = "zscore")
+#' winsorize(iris$Sepal.Length, threshold = 2, method = "zscore", robust = TRUE)
 #' winsorize(iris, threshold = 0.2)
 #' @inherit data_rename seealso
 #' @export
@@ -52,8 +54,8 @@ winsorize.data.frame <- function(data, threshold = 0.2, verbose = TRUE, robust =
 
 #' @rdname winsorize
 #' @export
-winsorize.numeric <- function(data, threshold = 0.2, verbose = TRUE, robust = FALSE, ...) {
-  if(robust == FALSE) {
+winsorize.numeric <- function(data, threshold = 0.2, verbose = TRUE, method = "percentile", robust = FALSE, ...) {
+  if(method == "percentile") {
 
     if (threshold < 0 || threshold > 0.5) {
       if (isTRUE(verbose)) {
@@ -75,7 +77,7 @@ winsorize.numeric <- function(data, threshold = 0.2, verbose = TRUE, robust = FA
     return(winval)
   }
 
-  if(robust == TRUE) {
+  if(method == "zscore") {
 
     if (threshold <= 0) {
       if (isTRUE(verbose)) {
@@ -84,11 +86,22 @@ winsorize.numeric <- function(data, threshold = 0.2, verbose = TRUE, robust = FA
       return(data)
     }
 
-    med <- stats::median(data, na.rm = TRUE)
-    y <- data - med
-    sc <- stats::mad(y, center = 0, na.rm = TRUE) * threshold
-    y[y > sc] <- sc
-    y[y < -sc] <- -sc
-    y + med
+    if(isTRUE(robust)) {
+      med <- stats::median(data, na.rm = TRUE)
+      y <- data - med
+      winval <- stats::mad(y, center = 0, na.rm = TRUE) * threshold
+      y[y > winval] <- winval
+      y[y < -winval] <- -winval
+      return(y + med)
+    }
+
+    if(isFALSE(robust)) {
+      m <- mean(data, na.rm = TRUE)
+      y <- data - m
+      winval <- stats::sd(y, na.rm = TRUE) * threshold
+      y[y > winval] <- winval
+      y[y < -winval] <- -winval
+      y + m
+    }
   }
 }
--- a/R/winsorize.R
+++ b/R/winsorize.R
@@ -65,8 +65,8 @@ winsorize.logical <- winsorize.factor
 #' @export
 winsorize.data.frame <- function(data, threshold = 0.2, method = "percentile", robust = FALSE,
                                  verbose = TRUE, ...) {
-  data <- lapply(data, winsorize, threshold = threshold, method = method, robust = robust, verbose = verbose)
-  as.data.frame(data)
+  data[] <- lapply(data, winsorize, threshold = threshold, method = method, robust = robust, verbose = verbose)
+  data
 }
 
 #' @rdname winsorize
@@ -78,7 +78,7 @@ winsorize.numeric <- function(data, threshold = 0.2, method = "percentile", robu
   if (method == "raw") {
     if (length(threshold) != 2L) {
       if (isTRUE(verbose)) {
-        warning("threshold must be of length 2 for lower and upper bound. Did not winsorize data.", call. = FALSE)
+        warning(insight::format_message("threshold must be of length 2 for lower and upper bound. Did not winsorize data."), call. = FALSE)
       }
       return(data)
     }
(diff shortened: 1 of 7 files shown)


R/winsorize.R
```diff
@@ -43,27 +45,50 @@ winsorize.character <- winsorize.factor
...
 
-  y <- sort(data)
-  n <- length(data)
-  ibot <- floor(threshold * n) + 1
-  itop <- length(data) - ibot + 1
-  xbot <- y[ibot]
-  xtop <- y[itop]
-  winval <- ifelse(data <= xbot, xbot, data)
-  ifelse(winval >= xtop, xtop, winval)
+  if(robust == TRUE) {
```
mattansb, line 78: This should probably be `isTRUE()`

mattansb:
There are two methods now:

1. Percentile based (`robust=FALSE`)
2. Median ± MAD (`robust=TRUE`)

However, both method can be considered "robust". 
Additionally, the classic Z-score (Mean ± SD) based method.


Maybe we can have
- `method = "percentile"`: x is between [threshold / 2, 1 - threshold /2]%
- `method = "zscore"`
  - `robust = FALSE`: x is between Mean ± threshold * SD
  - `robust = TRUE`: x is between Median ± threshold * MAD

What do you think?

R/winsorize.R
rempsyc: Learning everyday! I'm assuming this means that `robust == FALSE` (line 56) should be `isFALSE(robust)` too.

rempsyc:
My first (inexperienced) intuition was to set method to a specific option (e.g., `method = "percentile"`, `method = "mean"`, and `method = "median"` or `method = "mad"`). In our stats class, we are taught to use the MAD and it wasn't immediately clear to me that the trick to get that was to set `robust = TRUE` in other functions, though I was able to figure it out from the documentation. I'm just thinking that many of my peers don't have the habit to check the documentation like I do.

That said, I think we need to do as you suggest to stay consistent with the other `datawizard` functions. Also, it makes a lot of sense when you know a bit of stats. So I changed it accordingly. In this case, the `robust` argument only applies when `method = zscore` since as you mention `method = percentile` is robust already.

Another thing. In order to not let your previous work on #47 go to waste, I incorporated them in my PR before adding my changes. However, checks were already failing on that, so necessarily checks are failing on my modified version as well. I'm not familiar with automatic tests (yet) unfortunately, but I had a look at `tests/testthat/test-winsorization.R`. From what I can understand, the failed checks seem to be due to an error with one of the tests (i.e., the third one). So actually I just tried updating the expected message in the `regexp` call (from 1 to 0.5 since that's the modification you made to the function earlier) and it seems to have fixed it! :O

mattansb:
Okay, I've made some updates - I've made the code more modular (so now each method computes the thresholds and then those are treated). 
I've also added a "raw" method as per #47.

``` r
library(datawizard)
hist(iris$Sepal.Length, main = "Original data")
```



``` r
hist(winsorize(iris$Sepal.Length, threshold = 0.2),
     xlim = c(4, 8), main = "Percentile Winz")
```



``` r
hist(winsorize(iris$Sepal.Length, threshold = 1.5, method = "zscore"),
     xlim = c(4, 8), main = "Mean+-SD Winz")
```



``` r
hist(winsorize(iris$Sepal.Length, threshold = 1.5, method = "zscore", robust = TRUE),
     xlim = c(4, 8), main = "Median+-MAD Winz")
```



``` r
hist(winsorize(iris$Sepal.Length, threshold = c(5, 7.5), method = "raw"),
     xlim = c(4, 8), main = "Raw Thresholds")
```



<sup>Created on 2022-06-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

---

Note that as we currently have it set up, the percentile, the threshold argument defines the amount to windzorize from **each** tail. Is this intended? Desired? @IndrajeetPatil @DominiqueMakowski

mattansb:
Also, of course - @rempsyc great work!

rempsyc:
The only thing failing at this point is the `pkgdown` check from the vignette because `tidyr` is called but not part of the suggested packages, so I added it to `Suggests` in `DESCRIPTION` (plus made minor corrections to the documentation).

mattansb:
@IndrajeetPatil how should we avoid the `tidyr` issue?

IndrajeetPatil:
Great work, both of you! Thanks for that. 

As for _tidyr_, that's strange. The builds are not failing in the default branch without it, so how can they fail in the PR?!

Which _tidyr_ functions are we using in the vignette?

rempsyc:
`pkgdown` test doesn't fail anymore after adding `tydir` to `Suggests` in `DESCRIPTION` (https://github.com/easystats/datawizard/actions/runs/2563988526). So is the issue solved, or was the idea not to add it there for some reason?

IndrajeetPatil:
Not sure. I am currently traveling and don't have a laptop on me. I will have a look later.

@mattansb Feel free to squash and merge whenever you think this is ready for a merge.

rempsyc:
FYI I think the only `tidyr` function from the `standardize_data.Rmd` vignette (lines 76:86) is `tidyr::pivot_longer`. Do we have a `datawizard` equivalent? YES! And it's used just a bit further down in the vignette: `datawizard::data_to_long`. So what I did is just remove `tidyr` from `Suggests` and simply replaced `tidyr::pivot_longer` with `datawizard::data_to_long` in the vignette.

IndrajeetPatil:
Awesome! Thanks. ☺️

R/winsorize.R
```diff
@@ -43,27 +63,65 @@ winsorize.character <- winsorize.factor
 winsorize.logical <- winsorize.factor
 
 #' @export
-winsorize.data.frame <- function(data, threshold = 0.2, verbose = TRUE, ...) {
-  out <- sapply(data, winsorize, threshold = threshold, verbose = verbose)
-  as.data.frame(out)
+winsorize.data.frame <- function(data, threshold = 0.2, method = "percentile", robust = FALSE,
+                                 verbose = TRUE, ...) {
+  data <- lapply(data, winsorize, threshold = threshold, method = method, robust = robust, verbose = verbose)
+  as.data.frame(data)
```
strengejacke, line 69: I suggest using `data[] <- lapply...` and remove the next line with `as.data.frame()`.

R/winsorize.R
```diff
@@ -43,27 +63,65 @@ winsorize.character <- winsorize.factor
...
-      warning("'threshold' for winsorization must be a scalar between 0 and 1. Did not winsorize data.", call. = FALSE)
+winsorize.numeric <- function(data, threshold = 0.2, method = "percentile", robust = FALSE,
+                              verbose = TRUE, ...) {
+  method <- match.arg(method, choices = c("percentile", "zscore", "raw"))
+
+  if (method == "raw") {
+    if (length(threshold) != 2L) {
+      if (isTRUE(verbose)) {
+        warning("threshold must be of length 2 for lower and upper bound. Did not winsorize data.", call. = FALSE)
+      }
```
strengejacke, line 82: I suggest wrapping in `insight::format_message()`.

R/winsorize.R
```diff
@@ -43,27 +63,65 @@ winsorize.character <- winsorize.factor
...
+      return(data)
+    }
+  }
+
+  if(method == "percentile") {
+    if (threshold < 0 || threshold > 0.5) {
+      if (isTRUE(verbose)) {
+        warning("'threshold' for winsorization must be a scalar between 0 and 0.5. Did not winsorize data.", call. = FALSE)
+      }
```
strengejacke, line 91: Same here, use `warning(insight::format_message("..."), call.= FALSE)`.

R/winsorize.R
```diff
@@ -43,27 +63,65 @@ winsorize.character <- winsorize.factor
...
+
+    threshold <- c(y[ibot], y[itop])
+  }
+
+  if(method == "zscore") {
+
+    if (threshold <= 0) {
+      if (isTRUE(verbose)) {
+        warning("'threshold' for winsorization must be a scalar greater than 0. Did not winsorize data.", call. = FALSE)
+      }
```
strengejacke, line 108: See above.

tests/testthat/test-winsorization.R
```diff
@@ -11,11 +11,11 @@ test_that("with missing values", {
 test_that("winsorize: threshold must be between 0 and 1", {
   expect_warning(
     winsorize(sample(1:10, 5), threshold = -0.1),
-    regexp = "must be a scalar between 0 and 1"
+    regexp = "must be a scalar between 0 and 0.5"
   )
```
strengejacke, line 15: It might be that when `insight::format_message()` is used, a line break will by coincident just be inside this string, so matching does no longer work. Maybe just reduce the pattern to `0 and 0.5` or so?

R/winsorize.R
rempsyc: Done, but in that way it didn't return an output, so I had to add a line to return the dataframe anyway.

R/winsorize.R
rempsyc: Done

R/winsorize.R
rempsyc: Done

R/winsorize.R
rempsyc: Done

tests/testthat/test-winsorization.R
rempsyc: Actually this has been solved (essentially, we had modified the warning message within the function but not within the test checks. Harmonizing them fixed it)

strengejacke (approved):

rempsyc:
@mattansb , after working on `performance::check_outliers`, I noticed that instead of using "`zscore`" along with "`robust = TRUE`", they simply use "`zscore_robust`".

The intention of going with the `robust = TRUE` argument was to stay consistent with `datawizard::standardize`. However, for the former, there is only two methods: robust (median MAD) or not robust (mean SD), so it kind of makes sense.

However, now that we have several methods with `datawizard::winsorize` and that we had some confusion regarding `robust = TRUE` (since the argument currently applies to only one method), I wonder if we should go the `performance` way and just use `zscore_robust`. Perhaps it would be less confusing? By losing one parameter, I think we gain simplicity. What's the procedure in this case (if you agree), open a new PR?

bwiernik:
I say to keep robust and method as it is now

Issue #90, opened 2022-03-07
Vignette about data wrangling using `{datawizard}`


IndrajeetPatil:
Basically, an article demonstrating how the package (to a limited extent) can do similar things as `{dplyr}` + `{tidyr}` workflow.

IndrajeetPatil:
@etiennebacher Are you interested in taking this one on?

etiennebacher:
I can try to do something in the next days

Issue #59, opened 2022-01-28
Writing a JOSS paper


IndrajeetPatil:
Once the API is more stable (which should be the case in a few months), we should definitely do this.

IndrajeetPatil:
Submitting the paper today.

IndrajeetPatil:
The PDF for the most recent version of the paper can always be found here: 
https://github.com/easystats/datawizard/actions/workflows/draft-pdf.yml

DO NOT remove the related GitHub action because JOSS uses this action to build the paper.

IndrajeetPatil:
The paper is now under review: https://github.com/openjournals/joss-reviews/issues/4684

strengejacke:
Thanks for all the work on this paper!

PR #183, opened 2022-06-30
Add a vignette about equivalence with tidyverse

### LINKED ISSUE #130: Add some doc about transition from tidyverse?
I still use a lot the tidyverse because I know the main functions and how they work. I think that one thing missing in the docs is an article/table that shows equivalent functions between tidyverse and easystats ecosystems. This would do two things: 1) reduce the time spent looking for the equivalent of a function we use a lot with tidyverse, 2) highlight which functions of easystats don't have equivalent in tidyverse. I don't have something complicated in mind, maybe simply a table like:
| tidyverse | easystats| 
| ------------- |:-------------:| 
| `pull`      | `data_extract` | 
| `rename`      | `data_rename` |  
| `replace_na` | `convert_na_to` |
| ... | ... |    

To go further, this could be accompanied by some examples showing how to convert tidyverse workflows (just a few functions separated by a pipe) to easystats workflows. What do you think?

etiennebacher:
Closes #130

I put datawizard and tidyverse chunks side by side to easily compare the syntax, for example:

image

(I wanted to put different background colors to each chunk to distinguish between the two syntax, but apparently it is not yet supported by pkgdown (https://github.com/r-lib/downlit/issues/149) so I added comments in each chunk).


vignettes/basic_data_wrangling.Rmd
```diff
@@ -0,0 +1,518 @@
...
+  background-color: #d9f2e5;
+}
+```
+
+`datawizard` is a package whose aim is to make basic data wrangling easier than 
+with base R. Although it is inspired by the tidyverse combo `dplyr` + `tidyr`, 
+one of its main features is that it has very few dependencies: `stats` and `utils`
+(included in base R) and `insight`, which is the core package of the easystats 
+ecosystem. One drawback of this approach is that not all features of the 
+`tidyverse` packages are not supported and we will have to rely on base R, or on
```
strengejacke: This is ambiguous, because we don't rely on _poorman_ anywhere in the code.

vignettes/basic_data_wrangling.Rmd
```diff
@@ -0,0 +1,518 @@
...
+```
+
+
+
+
+## Selecting
+
+`data_select()` is the equivalent of `dplyr::select()`. The main difference 
+between these two functions is that `data_select()` uses two arguments (`select`
+and `exclude`) and requires quoted column names, while `dplyr::select()` accepts
```
strengejacke, line 118: No, there are many possibilities, including unquoted column names, but only for a single variable, I think (because we don't use `...` to capture comma-separated, unquoted names). But maybe this is too much to explain here, and for now, we can keep it like this.

strengejacke:
Nice draft! I like the structure of this vignette very much!

_pkgdown.yml
```diff
@@ -124,7 +124,12 @@ reference:
 
 articles:
 
+  - title: Basic Data Wrangling
+    navbar: ~
+    contents:
+    - basic_data_wrangling
```
IndrajeetPatil, line 130: If you name it "datawizard.Rmd", it will be rendered as "Get Started" vignette.

I think this vignette is the perfect candidate for that.

vignettes/basic_data_wrangling.Rmd
```diff
@@ -0,0 +1,518 @@
+---
+title: "Basic Data Wrangling"
+output: 
+  rmarkdown::html_vignette:
+    toc: true
+    fig_width: 10.08
+    fig_height: 6
```
IndrajeetPatil, line 7: Not needed.

```suggestion
```

vignettes/basic_data_wrangling.Rmd
```diff
@@ -0,0 +1,518 @@
...
+```{css, echo=FALSE, eval = TRUE}
+.datawizard, .datawizard > .sourceCode {
+  background-color: #e6e6ff;
+}
+.tidyverse, .tidyverse > .sourceCode {
+  background-color: #d9f2e5;
+}
+```
+
+`datawizard` is a package whose aim is to make basic data wrangling easier than
```
IndrajeetPatil, line 49:
```suggestion
`datawizard` package aims to make basic data wrangling easier than 
```

vignettes/basic_data_wrangling.Rmd
```diff
@@ -0,0 +1,518 @@
...
+.datawizard, .datawizard > .sourceCode {
+  background-color: #e6e6ff;
+}
+.tidyverse, .tidyverse > .sourceCode {
+  background-color: #d9f2e5;
+}
+```
+
+`datawizard` is a package whose aim is to make basic data wrangling easier than 
+with base R. Although it is inspired by the tidyverse combo `dplyr` + `tidyr`,
```
IndrajeetPatil, line 50: We should refer to packages as `{dplyr}`, and not as `dplyr`, to distinguish a package name from an object name.

vignettes/basic_data_wrangling.Rmd
```diff
@@ -0,0 +1,518 @@
...
+}
+.tidyverse, .tidyverse > .sourceCode {
+  background-color: #d9f2e5;
+}
+```
+
+`datawizard` is a package whose aim is to make basic data wrangling easier than 
+with base R. Although it is inspired by the tidyverse combo `dplyr` + `tidyr`, 
+one of its main features is that it has very few dependencies: `stats` and `utils`
+(included in base R) and `insight`, which is the core package of the easystats
```
IndrajeetPatil, line 52:
```suggestion
(included in base R) and `insight`, which is the core package of the _easystats_ 
```

Add a link to _easystats_ meta-package website.

IndrajeetPatil:
This is a fantastic start, @etiennebacher! Thanks for working on this. Lovely use of pandoc's fenced div!

I am going to make some additions and add a few edits, but overall this is a good structure. I think, once finalized, we can also melt this down a bit and convert it to a _JOSS_ paper (#59).

etiennebacher:
Thanks @IndrajeetPatil and @strengejacke! 

Before finishing this vignette, we need to improve `reshape_longer` and `reshape_wider` to match more precisely the tidyr implementations and outputs. I'll skim through datawizard later to see if I forgot some major functions, but maybe you already have some suggestions of additional functions to include?

_pkgdown.yml
IndrajeetPatil: I think, this vignette might not be the best fit for "Get Started" since it focuses exclusively on data wrangling functionality, and we also need to talk about data transformations. 

So we can keep the name as it is, and when we work on the JOSS paper, we can convert that paper to "Get Started" (https://github.com/easystats/easystats/issues/196).

vignettes/basic_data_wrangling.Rmd
DominiqueMakowski: Passing thought, maybe we should add in our style guide in easystats the fact that we prefer quoted names rather than non-standard eval, because it's easier to program with (in easystats, as opposed to the tidyverse), because it's flexible, it makes more sense to distinguish names from objects, it's a much more standard programming good practice, and that in general NSE is an eldritch invention

IndrajeetPatil:
> but maybe you already have some suggestions of additional functions to include

I have added a couple of TODOs for this in the vignette. I will keep updating them as I think about more.

IndrajeetPatil:
@etiennebacher I am changing the targeted issue for this PR to #130 because that's what the PR is targeting.

For #90, I had a very different kind of vignette in mind. 
Specifically, one where we start with a messy dataset and achieve the desired result using data wrangling functions. Something similar to [this](https://tidyr.tidyverse.org/articles/pivot.html) article.

etiennebacher:
> For #90, I had a very different kind of vignette in mind. Specifically, one where we start with a messy dataset and achieve the desired result using data wrangling functions. Something similar to [this](https://tidyr.tidyverse.org/articles/pivot.html) article.

I see, I think a good example of a messy dataset could be to download a csv file from the World bank catalogue (not using the package `WDI` because it provides already cleaned data). We could use:
* filtering to keep countries only (remove aggregates)
* pivoting because it's one year per col
* selecting, renaming and relocating for diverse things

IndrajeetPatil:
Let's discuss this in #90. 

But feel free to modify this PR to address my TODOs for now. I won't be making any changes right now.

vignettes/basic_data_wrangling.Rmd
etiennebacher: done

etiennebacher:
I just discovered that `data_extract()` has an argument `name` that is used when we extract a single column, but I don't see the point of this. Anyone has a usecase for this?
``` r
library(datawizard)
data_extract(mtcars, "cyl", name = gear)
#> 4 4 4 3 3 3 3 4 4 4 4 3 3 3 3 3 3 4 4 4 3 3 3 3 3 4 5 5 5 5 5 4 
#> 6 6 4 6 8 6 8 4 4 6 6 8 8 8 8 8 8 4 4 4 4 8 8 8 8 4 4 4 8 6 8 4
```

<sup>Created on 2022-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

IndrajeetPatil:
I'd leave it be.

I think it can be used as a kind of look-up table:

``` r
library(datawizard)

# find brainwt values for animals that have no vorosity attached to them
x <- data_extract(ggplot2::msleep, "brainwt", name = "vore")
x[is.na(names(x))]
#>    <NA>    <NA>    <NA>    <NA>    <NA>    <NA>    <NA> 
#>      NA 0.00240      NA 0.01140 0.02100 0.00300 0.00033
```

<sup>Created on 2022-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

vignettes/tidyverse_translation.Rmd
```diff
@@ -275,7 +275,45 @@ You can find a list of all the select helpers with `?data_select`.
 
 ## Extracting {#extracting}
 
-<!-- TODO: data_extract and dplyr::pull -->
+<!-- TODO: improve motivation for data_extract -->
```
IndrajeetPatil, line 278: TBH, I don't think we need to motivate these functions here, since this vignette is just to set up translation with _tidyverse_. 

I had initially added sentences like these, but have already removed them.

strengejacke:
> I just discovered that `data_extract()` has an argument `name` that is used when we extract a single column, but I don't see the point of this. Anyone has a usecase for this?
> 
> ```r
> library(datawizard)
> data_extract(mtcars, "cyl", name = gear)
> #> 4 4 4 3 3 3 3 4 4 4 4 3 3 3 3 3 3 4 4 4 3 3 3 3 3 4 5 5 5 5 5 4 
> #> 6 6 4 6 8 6 8 4 4 6 6 8 8 8 8 8 8 4 4 4 4 8 8 8 8 4 4 4 8 6 8 4
> ```
> 
> Created on 2022-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)

@DominiqueMakowski @bwiernik I think one of you two had the idea (and a use case in mind)

bwiernik:
It sets the names of the vector elements to another column. 

```r
data.frame(
  key = letters,
  value = 1:26
) |>
  data_extract("value", name = key)
```

DominiqueMakowski:
> I think one of you two had the idea

not that i can think of

etiennebacher:
> It sets the names of the vector elements to another column.

Yes but I was just wondering if you had a real life example where this was needed (I'm not arguing to remove this argument, just curious about when it should be used)

bwiernik:
it's analogous to deframe()

vignettes/tidyverse_translation.Rmd
```diff
@@ -0,0 +1,771 @@
...
+key functions in *datawizard* for data wrangling:
+
+Function           | Operation                                           |
+------------------ | --------------------------------------------------- |
+`data_filter()`    | [to select only certain *observations*](#filtering) |
+`data_select()`    | [to select only a few *attributes*](#selecting)     |
+`data_extract()`   | [to extract a single *attribute*](#extracting)      |
+`data_rename()`    | [to rename attributes](#renaming)                   |
+`data_relocate()`  | [to reorder a data frame](#relocating)              |
+`data_rotate()`    | to rotate (or transpose) a data frame               |
```
etiennebacher, line 87: I don't think this has an equivalent in the tidyverse, right?

etiennebacher: ok I let it there

IndrajeetPatil:
I don't have strong opinions on that, but I personally prefer showcasing data wrangling functions having the `data_*()` prefix because it's an easy pattern for the users to remember.

IndrajeetPatil:
I also feel that we should remove the `data_*()` prefix from statistical transformation function names.

etiennebacher:
> I also feel that we should remove the `data_*()` prefix from statistical transformation function names.

@IndrajeetPatil I created a new issue based on this, since we don't show any statistical transformation in this vignette

IndrajeetPatil:
@etiennebacher I'd like to merge this soon. I think most of the `TODO`s have been addressed, so this should be ready.

We can keep updating it as we add more tidyverse equivalents (e.g. `arrange()` <-> `data_arrange()`).

IndrajeetPatil (approved):

IndrajeetPatil:
Merci beaucoup, Étienne!

PR #185, opened 2022-07-01
Deprecate before and after in data_relocate

etiennebacher:
This PR deprecates `before` and `after` in `data_relocate()` to use `.before` and `.after` instead, like in `dplyr::relocate()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling
-Version: 0.4.1.6
+Version: 0.4.1.7
 Authors@R: c(
     person("Dominique", "Makowski", , "dom.makowski@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0001-5375-9967", Twitter = "@Dom_Makowski")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,9 @@ CHANGES
 
 * `data_partition()` now allows to create multiple partitions from the data,
   returning multiple training and a remaining test set.
+  
+* Arguments `before` and `after` are deprecated in `data_relocate()` for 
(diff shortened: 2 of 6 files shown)


strengejacke:
@bwiernik @IndrajeetPatil @DominiqueMakowski I know we were trying to have similar / same argument names in the pivot-functions, but do we want to apply this to as many functions as possible in _datawizard_?

I personally don't think we should do this if we leave our code-style guidelines (though, we have not many specifications regarding _naming of arguments_).

IndrajeetPatil:
I agree with Daniel here. I don't think we need to introduce two different styles of arguments just to be consistent with the _tidyverse_, because the cost of that consistency is more confusion for _easystats_ users because they will have to keep guessing which arguments have `.` and which don't.

For example, the _dplyr_ function has `.data`, but the equivalent parameter is called `data` in _datawizard_ function. Should we try to be consistent with that parameter name as well? I don't think so.

``` r
names(formals(dplyr::relocate))
#> [1] ".data"   "..."     ".before" ".after"
```

<sup>Created on 2022-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Better to let the users form a predictable mental model for what parameter names are going to: always without a `.`.

etiennebacher:
Sure, makes sense

bwiernik:
We might consider catching tidyverse-style names and passing them to their corresponding arguments?

DominiqueMakowski:
> We might consider catching tidyverse-style names and passing them to their corresponding arguments?

Is there a way to do that in an elegant manner? Using some internal function for instance that would assign all dotted args to non-dotted?

Otherwise, I agree to stick with our easystats naming rather than chasing tidyverse

bwiernik:
My understanding is that the motivation of .before/.after in tidyverse is because "before" might more likely be a name someone would actually want to use in their data, so having the dot prevents that more common collision

PR #187, opened 2022-07-04
Fix formatting issue in vignette

etiennebacher:
In vignette "standardize data":
* before

image

* after 

image

### DIFF
--- a/vignettes/standardize_data.Rmd
+++ b/vignettes/standardize_data.Rmd
@@ -15,7 +15,6 @@ vignette: >
 library(knitr)
 options(knitr.kable.NA = "")
 knitr::opts_chunk$set(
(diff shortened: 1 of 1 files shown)


vignettes/standardize_data.Rmd
```diff
@@ -15,7 +15,6 @@ vignette: >
 library(knitr)
 options(knitr.kable.NA = "")
 knitr::opts_chunk$set(
-  comment = ">",
```
IndrajeetPatil, line 18: Shouldn't we add `comment = "#>",` instead?

That would also remove ANSI escape sequences from showing up, IINM.

etiennebacher: done

Issue #200, opened 2022-07-13
Cache GitHub actions 


bwiernik:
It looks like the README render action is building several large dependencies (eg, psych) each time. Can we add the cache action to speed that up? https://github.com/easystats/datawizard/runs/7319904358?check_suite_focus=true

@IndrajeetPatil

IndrajeetPatil:
I am surprised that this is not already the case. We use `setup-r-dependencies` workflow to download dependencies:

https://github.com/easystats/datawizard/blob/9b709a4bf86f97923a0954810ce5747510779299/.github/workflows/render-readme.yml#L30

And, by default, this workflow caches deps:

https://github.com/r-lib/actions/blob/4e1feaf90520ec1215d1882fdddfe3411c08e492/setup-r-dependencies/action.yaml#L5-L8

At any rate, I am trying something in https://github.com/easystats/datawizard/commit/9b709a4bf86f97923a0954810ce5747510779299. Let's see if it works.

IndrajeetPatil:
I don't get it.

If you look at the latest render README GHA [log](https://github.com/easystats/datawizard/runs/7361925059?check_suite_focus=true), it detects cache, but still goes on to install everything from scratch and take over an hour!

<img width="1028" alt="Screenshot 2022-07-16 at 22 08 41" src="https://user-images.githubusercontent.com/11330453/179370581-64fa7f19-fd10-48b3-be02-bff615e17fde.png">

While for R CMD Check [GHA](https://github.com/easystats/datawizard/runs/7361925211?check_suite_focus=true), it takes just a minute!

<img width="1029" alt="Screenshot 2022-07-16 at 22 10 37" src="https://user-images.githubusercontent.com/11330453/179370598-c2e1c95a-288c-4e9f-8f3d-f6dc76952e3b.png">

Why the difference?! @etiennebacher, do you have any idea about this?

etiennebacher:
The only difference I see on the screenshots is that render-readme uses ubuntu while r-cmdcheck uses macos. Maybe using macos on render-readme will fix the caching behavior?

IndrajeetPatil:
I am not sure why that would make a difference because we also do R CMD Check on Ubuntu and the installation time is again just over a minute.

<img width="1027" alt="Screenshot 2022-07-16 at 22 54 16" src="https://user-images.githubusercontent.com/11330453/179371684-8623bee8-c092-478f-8416-5b3c52678100.png">

IndrajeetPatil:
[...]

PR #201, opened 2022-07-16
use macos for render-readme

etiennebacher:
related to #200


etiennebacher:
@IndrajeetPatil Looks like this worked, render-readme took 3 min only

IndrajeetPatil (approved):

IndrajeetPatil:
Strange that this works on Mac, but not on Ubuntu.

At any rate, we will need to make the same change across all repos since we are using the same GHA everywhere.

PR #205, opened 2022-07-22
Rename `to_numeric()`

IndrajeetPatil:
For now.

Later, this will be removed and incorporated into the new `to_numeric()`.


IndrajeetPatil:
@bwiernik This PR broke _correlation_, but I've already fixed it in [this](https://github.com/easystats/correlation/pull/244) PR.

So, once I updated _datawizard_ on CRAN, you will also need to update _correlation_ (https://github.com/easystats/easystats/issues/256).

PR #212, opened 2022-08-06
revdepcheck

IndrajeetPatil:
@strengejacke Lots of breakages 🙈

### DIFF
--- a/revdep/README.md
+++ b/revdep/README.md
@@ -1,25 +1,36 @@
 # Platform
 
-|field    |value                                                               |
-|:--------|:-------------------------------------------------------------------|
-|version  |R version 4.1.2 (2021-11-01)                                        |
-|os       |Windows 10 x64 (build 22000)                                        |
-|system   |x86_64, mingw32                                                     |
-|ui       |RStudio                                                             |
-|language |(EN)                                                                |
-|collate  |English_United Kingdom.1252                                         |
-|ctype    |English_United Kingdom.1252                                         |
-|tz       |Europe/Berlin                                                       |
-|date     |2022-02-25                                                          |
-|rstudio  |2022.06.0-daily+135 Spotted Wakerobin (desktop)                     |
-|pandoc   |2.17.1.1 @ C:/Program Files/RStudio/bin/quarto/bin/ (via rmarkdown) |
+|field    |value                                     |
+|:--------|:-----------------------------------------|
+|version  |R version 4.2.1 (2022-06-23)              |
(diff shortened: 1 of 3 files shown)


strengejacke:
Some of the breakages come from `data_to_long()` when the former `cols` argument is used inside a function. That just needs to be replaced by `select` (else, NSE doesn't work properly, as we see from all the breakages). I think we have indeed fixed everything, and I currently don't see a way to avoid this. :-/

strengejacke:
@mattansb are the issues for _effectsize_ fixed, or is there something we can do on _datawizard_'s side to avoid the revdep issue?

strengejacke:
- _statsExpressions_ breaks due to _effectsize_
- _sjPlot_ can probably submitted now, so we shouldn't have breakage there
- _see_, _report_, _parameters_ and _correlation_ should already be fixed.

Maybe we wait for _sjPlot_ to be submitted, if we can do something in regard to _effectsize_ and then submit?

mattansb:
I just ran effectsize tests with the newest datawizrd, and didn't get those errors, so... 🤷

strengejacke:
The question is whether this issue, occurring for the current *CRAN* version of _effectsize_, is something we need to fix in _datawizard_, or if this is a necessary and expected breakage.

```
      Running ‘spelling.R’
      Running ‘testthat.R’
     ERROR
    Running the tests in ‘tests/testthat.R’ failed.
    Last 13 lines of output:
       7.       └─effectsize (local) FUN(newX[, i], ...)
       8.         └─datawizard::ranktransform(x, method = "average", ..., verbose = FALSE)
      ── Error (test-rankES.R:76:5): kendalls_w ──────────────────────────────────────
      Error in `UseMethod("ranktransform")`: no applicable method for 'ranktransform' applied to an object of class "character"
      Backtrace:
    ...
          ▆
       1. └─effectsize::kendalls_w(M1) at test-rankES.R:76:4
       2.   └─effectsize:::.kendalls_w(data, verbose = verbose)
       3.     └─base::apply(data, 1, .safe_ranktransform, verbose = verbose)
       4.       └─effectsize (local) FUN(newX[, i], ...)
       5.         └─datawizard::ranktransform(x, method = "average", ..., verbose = FALSE)
      
      [ FAIL 3 | WARN 3 | SKIP 11 | PASS 494 ]
      Error: Test failures
      Execution halted
```

mattansb:
I think that was solved previously.
But I can send a patch fix to CRAN if needed.

PR #235, opened 2022-08-24
Rename `change_code()` to `recode_values()`

### LINKED ISSUE #233: Re-rename `change_code()`?
Before we finalize this, we should definitely decide on the new function names. Mostly, I'm not quite satisfied with `change_code()`. `recode` is a verb, and everyone would expect such a function would recode old into new values. in `change_code()`, the verb is `change`, and what can we expect if we change a code? What we actually _change_ when _recoding_ are _values_ (or factor levels, but values is maybe more generic). What do you think about `change_values()`? or maybe `recode_values()`, or `recode_variables()`.

_Originally posted by @strengejacke in https://github.com/easystats/datawizard/issues/87#issuecomment-1200449988_

IndrajeetPatil:
- Closes #233
- style
- rename files to reflect new function names


strengejacke:
Do we still have the alias `change_code()` in case we need it for the paper submission? (else, we probably still could keep the alias for one release cycle, including a deprecation warning)

IndrajeetPatil:
Yes

https://github.com/easystats/datawizard/blob/01081172e167838f76965427b5bb582a71552aa1/NAMESPACE#L159

strengejacke (approved):

Issue #232, opened 2022-08-24
Add more summary statistics measures from `{sjstats}`


IndrajeetPatil:
Maybe also some of the [summary statistics](https://strengejacke.github.io/sjstats/reference/index.html#section-other-summary-statistics) from sjstats can be implemented in datawizard?

_Originally posted by @strengejacke in https://github.com/easystats/datawizard/issues/12#issuecomment-854410266_

bwiernik:
My preference would be to implement a set of descriptive functions with general names (e.g., `distribution_mean`, `distribution_sd`, `distribution_mad`, `distribution_cv`) that consistently support a common set of useful arguments:

1. `weights`
2. `trim`
3. `method` (e.g., to compute maximum likelihood, unbiased, or robust versions of a stat).

and perhaps other arguments. 

Small point: I really hate how "standard error" is commonly used to refer specifically to the SE of the mean, so if we added something like that, I would prefer it be something like `distribution_sem()` or `distribution_se_mean()`.

etiennebacher:
Where are we with this? Currently, the "Data properties" section contains these functions:

![image](https://user-images.githubusercontent.com/52219252/225239602-9797f794-7215-419f-ba27-495ee9870270.png)

It would be useful to have a clear list of functions to add or import from other packages.

elinw:
I just created these exact functions (with the same names) in a package for a course. It would be really great to have these.
The one thing I did is have a parameter for whether `to_numeric = FALSE` that allows you to specify converting non numeric data (e.g. ordered factors or numbers stored as characters).

etiennebacher:
> I just created these exact functions (with the same names) in a package for a course. It would be really great to have these.

Which functions are you referring to?

PR #236, opened 2022-08-28
Add coefficient of variation function

bwiernik:
Related to https://github.com/easystats/datawizard/issues/232, https://github.com/easystats/performance/issues/433


IndrajeetPatil:
Thanks!

`R CMD check` seems to be failing because of docs issue.

Also, can you please add some tests?

strengejacke:
This currently cannot work. See first example: `coef_var(1:10)`
In this case, both `mu` and `sigma` are `NULL`, and the first line in the function that is executed is `out <- sigma / mu` (which will be an empty vector as result).

bwiernik:
Didn't finish writing/testing yesterday. You're right Daniel, the individual tests for sigma and mu should be is.null() not !is.null()

bwiernik:
@IndrajeetPatil @strengejacke Anyone know how to get Roxygen2 to put the `...` argument last in the list?

R/descriptives.R
```diff
@@ -0,0 +1,161 @@
...
+#' coef_var(mu = 10, sigma = 20, method = "unbiased", n = 30)
+#' cv(1:10)
+coef_var <- function(x, ...) {
+  UseMethod("coef_var")
+}
+
+#' @name cv
+#' @rdname coef_var
+#' @export
+cv <- coef_var
```
IndrajeetPatil, line 53: I am wondering if we should support this alias. It is quite non-descriptive, and can be easily confused with cross-validation.

R/descriptives.R
```diff
@@ -0,0 +1,161 @@
...
+
+#' @name distribution_cv
+#' @rdname coef_var
+#' @export
+distribution_cv <- coef_var
```
IndrajeetPatil, line 58: If we remove alias `cv`, then this can be renamed to `distribution_ coef_var()`. It'll be verbose, but auto-completion will take care of it.

R/descriptives.R
bwiernik: Yeah I'm torn. "CV" is by far the most common way to write the stat name (analogous to "SD"), but yeah it's not that descriptive and is easily confused with cross validation. 

I guess I lean toward not including. Thoughts @DominiqueMakowski @mattansb @strengejacke ?

IndrajeetPatil: @DominiqueMakowski @mattansb @strengejacke @etiennebacher WDYT?

etiennebacher: `coef_var` is explicit and short enough to me, and I agree that `cv` can be confused with cross validation

R/descriptives.R
```diff
@@ -0,0 +1,154 @@
...
+
+
+# descriptives helpers
+
+.drop_na <- function(x) {
+  x[!is.na(x)]
+}
+
+.trim_values <- function(x, trim = 0, n = NULL, weights = NULL) {
+  # TODO: Support weights
```
IndrajeetPatil, line 139: @bwiernik Can this already be merged, or do you also want to work in this PR on supporting weights here?

mattansb:
I would like to add my suggestions please before merging (:

R/descriptives.R
```diff
@@ -82,13 +82,38 @@ coef_var.default <- function(x, verbose = TRUE, ...) {
...
+#' possible value), and that ratios between them are interpretable For example,
+#' how many sandwiches have I eaten this week? 0 means "none" and 20 sandwiches
+#' is 4 times more than 5 sandwiches. If I were to center the number of
+#' sandwiches, it will no longer be on a ratio scale (0 is no "none" it is the
+#' mean, and the ratio between 4 and -2 is not meaningful). Scaling a ratio
+#' scale still results in a ratio scale. So I can re define "how many half
+#' sandwiches did I eat this week ( = sandwiches * 0.5) and 0 would still mean
+#' "none", and 20 half-sandwiches is still 4 times more than 5 half-sandwiches.
+#'
+#' This means that CV is **NOT** invariance to shifting, but it is to scaling:
```
IndrajeetPatil, line 97:
```suggestion
#' This means that CV is **NOT** invariance to shifting, but it is to scaling:

```

mattansb (approved):

R/descriptives.R
strengejacke: agree, we should remove this alias.

R/descriptives.R
strengejacke: We should at least print a message if `weights` has no functionality yet.

strengejacke:
> @IndrajeetPatil @strengejacke Anyone know how to get Roxygen2 to put the `...` argument last in the list?

I'm not sure this is possible, I think it follows the "help file" logic, where you see all arguments of the first function in "Usage", then remaining arguments from following functions.

strengejacke (approved):

PR #240, opened 2022-09-01
add regex to `data_addprefix` for correct use in `performance::check_outliers`

rempsyc:
Forgot this one last time so it's creating a bug in `performance::check_outliers`


strengejacke (approved):

IndrajeetPatil:
Thanks!

PR #249, opened 2022-09-03
fix tests with R 3.6

### LINKED ISSUE #244: Fix tests for `R 3.6`
Since we support this version, tests shouldn't fail.

https://github.com/easystats/datawizard/runs/8151059594?check_suite_focus=true

etiennebacher:
Close #244. Almost all failures were caused by the fact that `data.frame()` used `stringsAsFactors = TRUE` by default before R 4.0 and `FALSE` after R 4.0


IndrajeetPatil (approved):
Thanks, @etiennebacher!

PR #252, opened 2022-09-05
data_to_long: improve speed with names_pattern

etiennebacher:
Improve speed when using `names_pattern`. 

Current results:
``` r
library(tidyr)
library(datawizard)

bench::mark(
  reshape_longer = who %>%
    reshape_longer(
      select = 5:60,
      names_to = c("diagnosis", "gender", "age"),
      names_pattern = "new_?(.*)_(.)(.*)",
      values_to = "count"
    )
)
#> # A tibble: 1 × 6
#>   expression          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 reshape_longer    3.56s    3.56s     0.281     768MB     4.21
```

New results:
``` r
library(tidyr)
library(datawizard)

bench::mark(
  reshape_longer = who %>%
    reshape_longer(
      select = 5:60,
      names_to = c("diagnosis", "gender", "age"),
      names_pattern = "new_?(.*)_(.)(.*)",
      values_to = "count"
    )
)
#> # A tibble: 1 × 6
#>   expression          min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>     <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 reshape_longer    1.69s    1.69s     0.592     367MB     7.11
```

### DIFF
--- a/R/data_reshape.R
+++ b/R/data_reshape.R
@@ -165,7 +165,7 @@ data_to_long <- function(data,
 
   # Cleaning --------------------------
   # Sort the dataframe (to match pivot_longer's output)
-  long <- data_arrange(long, c("_Row", names_to_2))
+  long <- long[do.call(order, long[, c("_Row", names_to_2)]), ]
 
   # Remove or rename the row index
   if (is.null(rows_to)) {
--- a/R/data_reshape.R
+++ b/R/data_reshape.R
@@ -165,7 +165,7 @@ data_to_long <- function(data,
 
   # Cleaning --------------------------
   # Sort the dataframe (to match pivot_longer's output)
-  long <- long[do.call(order, long[, c("_Row", names_to_2)]), ]
+  long <- data_arrange(long, c("_Row", names_to_2))
 
   # Remove or rename the row index
   if (is.null(rows_to)) {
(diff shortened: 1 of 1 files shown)


R/data_reshape.R
```diff
@@ -165,7 +165,7 @@ data_to_long <- function(data,
 
   # Cleaning --------------------------
   # Sort the dataframe (to match pivot_longer's output)
-  long <- long[do.call(order, long[, c("_Row", names_to_2)]), ]
+  long <- data_arrange(long, c("_Row", names_to_2))
```
strengejacke, line 168: w/o this change, it's even a bit faster, or was there a particular reason for calling `data_arrange()`?

etiennebacher: ah I thought it was equivalent in speed but more readable, but I should revert it if do.call is faster

strengejacke: My benchmark showed 1.62 vs 1.52 seconds, not much, but maybe still?

etiennebacher: Yeah on my side I win almost 0.5 sec (it changes at each bench::mark run but still do.call is faster. I'll revert the change

strengejacke: I think the minimal overhead is regex usage:

```r
  desc <- select[grepl("^-", select)]
  desc <- gsub("^-", "", desc)
  select <- gsub("^-", "", select)
```

strengejacke (approved):

R/data_reshape.R
etiennebacher: using `fixed = TRUE` barely reduces the time for `data_arrange`, the overhead probably comes from all the checks in `data_arrange`

PR #254, opened 2022-09-05
Add some tests for demean and data_merge

etiennebacher:
The default behavior of `data_merge()` when `by` columns are not present in both datasets is to warn the user and use all columns that are in both datasets to merge them (even if they are not specified in `by` originally).

This is really odd to me, I don't see when this behavior would be desired. I think an error would be more appropriate (which is the behavior in merge functions of dplyr btw).

### DIFF
--- a/R/data_merge.R
+++ b/R/data_merge.R
@@ -220,16 +220,23 @@ data_merge.data.frame <- function(x, y, join = "left", by = NULL, id = NULL, ver
     # If not all column names specified in "by" are present, yield warning
     # and use all shared column names
     if (!all(by %in% colnames(x)) || !all(by %in% colnames(y))) {
+      missing_in_x <- setdiff(by, colnames(x))
+      missing_in_y <- setdiff(by, colnames(y))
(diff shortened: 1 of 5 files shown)


R/data_merge.R
```diff
@@ -220,16 +220,23 @@ data_merge.data.frame <- function(x, y, join = "left", by = NULL, id = NULL, ver
     # If not all column names specified in "by" are present, yield warning
     # and use all shared column names
     if (!all(by %in% colnames(x)) || !all(by %in% colnames(y))) {
+      missing_in_x <- setdiff(by, colnames(x))
+      missing_in_y <- setdiff(by, colnames(y))
+      stop_message <- c("Not all columns specified in `by` were found in the data frames.",
+      if (length(missing_in_x) > 0) {
+        paste0("Following columns are in `by` but absent in `x`: ", text_concatenate(missing_in_x))
```
strengejacke, line 227: You don't add the result from paste to stopmessage.

strengejacke: True, I missed the `c()` and comma.

strengejacke (approved):

Issue #180, opened 2022-06-28
select helpers don't work in loops


etiennebacher:
``` r
library(datawizard)
iris <- head(iris)

iris |>
  data_relocate(select = starts_with("Sepal"), after = -1)
#>   Petal.Length Petal.Width Species Sepal.Length Sepal.Width
#> 1          1.4         0.2  setosa          5.1         3.5
#> 2          1.4         0.2  setosa          4.9         3.0
#> 3          1.3         0.2  setosa          4.7         3.2
#> 4          1.5         0.2  setosa          4.6         3.1
#> 5          1.4         0.2  setosa          5.0         3.6
#> 6          1.7         0.4  setosa          5.4         3.9

for (i in "Sepal") {
  iris |>
    data_relocate(select = starts_with(i), after = -1) |>
    print()
}
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1          5.1         3.5          1.4         0.2  setosa
#> 2          4.9         3.0          1.4         0.2  setosa
#> 3          4.7         3.2          1.3         0.2  setosa
#> 4          4.6         3.1          1.5         0.2  setosa
#> 5          5.0         3.6          1.4         0.2  setosa
#> 6          5.4         3.9          1.7         0.4  setosa


iris |>
  data_relocate(select = ends_with("Length"), after = -1)
#>   Sepal.Width Petal.Width Species Sepal.Length Petal.Length
#> 1         3.5         0.2  setosa          5.1          1.4
#> 2         3.0         0.2  setosa          4.9          1.4
#> 3         3.2         0.2  setosa          4.7          1.3
#> 4         3.1         0.2  setosa          4.6          1.5
#> 5         3.6         0.2  setosa          5.0          1.4
#> 6         3.9         0.4  setosa          5.4          1.7

for (i in "Length") {
  iris |>
    data_relocate(select = ends_with(i), after = -1) |>
    print()
}
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1          5.1         3.5          1.4         0.2  setosa
#> 2          4.9         3.0          1.4         0.2  setosa
#> 3          4.7         3.2          1.3         0.2  setosa
#> 4          4.6         3.1          1.5         0.2  setosa
#> 5          5.0         3.6          1.4         0.2  setosa
#> 6          5.4         3.9          1.7         0.4  setosa
```

[...]

PR #255, opened 2022-09-06
.select_nse: allow passing arguments through functions and loops

etiennebacher:
Outdated message, better to see the examples below.

<details>
<summary>See outdated message </summary>

I wanted to investigate on #180 (how to pass `select` and `exclude` through functions arguments or loop values) so I started rewriting `.select_nse()` with that in mind, hoping that I could have a clean implementation. It didn't go according to plan, due to the huge number of arguments that `select` and `exclude` accept. There are a lot of `if` conditions which correspond to many corner cases. Nevertheless, the new implementation seems to work with loops and functions (and passes previous tests, at least locally).

Here are just a few examples from #180 and from [the docs](https://easystats.github.io/datawizard/reference/find_columns.html#details).

### Functions 

``` r
library(datawizard)

foo <- function(data) {
  i <- "Sep"
  find_columns(data, select = starts_with(i))
}
foo(iris)
#> [1] "Sepal.Length" "Sepal.Width"
```

It also works if `i` is defined in the global environment (the `i` that is defined in the function is correctly used):

``` r
i <- "Petal"
foo <- function(data) {
  i <- "Sep"
  find_columns(data, select = starts_with(i))
}
foo(iris)
#> [1] "Sepal.Length" "Sepal.Width"
```

Also possible to pass the whole `starts_with()` as function argument:

``` r
foo <- function(data, arg) {
  find_columns(data, select = arg)
}
foo(iris, starts_with("Sep"))
#> Warning in get0(x, envir = env, inherits = inherits, ifnotfound = myObj):
#> redémarrage de l'évaluation d'une promesse interrompue
#> Warning in eval(select): redémarrage de l'évaluation d'une promesse interrompue
#> Warning in get0(x, envir = env, inherits = inherits, ifnotfound = myObj):
#> redémarrage de l'évaluation d'une promesse interrompue
#> [1] "Sepal.Length" "Sepal.Width"
```

As you can see, the use of `dynGet()` in the code created some messages that I don't how to remove (yet, hopefully).

### Loops
``` r
library(datawizard)

for (i in "Sepal") {
  head(iris) |>
    data_relocate(select = starts_with(i), after = -1) |>
    print()
}
#>   Petal.Length Petal.Width Species Sepal.Length Sepal.Width
#> 1          1.4         0.2  setosa          5.1         3.5
#> 2          1.4         0.2  setosa          4.9         3.0
#> 3          1.3         0.2  setosa          4.7         3.2
#> 4          1.5         0.2  setosa          4.6         3.1
#> 5          1.4         0.2  setosa          5.0         3.6
#> 6          1.7         0.4  setosa          5.4         3.9

for (i in "Length") {
  head(iris) |>
    data_relocate(select = ends_with(i), after = -1) |>
    print()
}
#>   Sepal.Width Petal.Width Species Sepal.Length Petal.Length
#> 1         3.5         0.2  setosa          5.1          1.4
#> 2         3.0         0.2  setosa          4.9          1.4
#> 3         3.2         0.2  setosa          4.7          1.3
#> 4         3.1         0.2  setosa          4.6          1.5
#> 5         3.6         0.2  setosa          5.0          1.4
#> 6         3.9         0.4  setosa          5.4          1.7
```
What do you think about this? It would be great if you had more examples to test this new behaviour. Maybe it is possible to keep the original structure of `.select_nse()` and to implement some of my code to allow passing args through functions and loops. I'll work on that later this week if needed. 

It's also possible that this new code is not robust enough, especially with all the `parent.frame()` and `dynGet()` calls. 

</details>


tests/testthat/test-find_columns.R
```diff
@@ -34,11 +34,6 @@ test_that("find_columns works as expected", {
     c("gear", "cyl")
   )
 
-  expect_equal(
-    find_columns(iris, Spec),
-    "Species"
-  )
-
```
etiennebacher, line 41: I don't think it's a good idea to support something like this, which is why I deleted it. It is safer to use "Spec" and `regex = TRUE`

R/select_helpers.R
```diff
@@ -1,274 +1,409 @@
...
-      stop(insight::format_message(
-        "The `data` argument must be a data frame, or an object that can be coerced to a data frame."
-      ), call. = FALSE)
+  .check_data(data)
+  cols <- names(data)
+
+  # avoid conflicts for starts_with(), etc.
+  # (has to be in this order, otherwise error because e.g tidyselect is imported
+  # in tidyr)
+  conflicting_packages <- .conflicting_packages(c("poorman", "tidyr", "dplyr", "tidyselect"))
```
strengejacke, line 11: I'm not sure we need all packages here, I didn't have any problems with tidyr, dplyr, etc. Or have you found situations where this errors?

strengejacke:
I think this PR is difficult to check, thus I would say we need to trust our tests here. :-)

etiennebacher:
No need for a review now, this PR is not ready and there are two tests failing

etiennebacher:
Tests are passing (except those with Ubuntu and R >= 4.2.0 because of [this](https://github.com/easystats/easystats/issues/234#issuecomment-1238484373)). Todo before this is ready for review:
- tests:
  - [ ] check the tests I modified 
  - [ ] add tests (or examples) for the new behavior when `dplyr`/`tidyr` are loaded to see if it messes up with the select helpers functions
  - [ ] add tests for some functions, e.g `data_relocate()`, that uses loops or functions for select helpers? or is it duplicated with the tests added for `get_columns()`?
  - [ ] reverse dependency checks to see if it affects the other packages of the ecosystem?
- docs:
  - [ ] remove the part in the docs where it is mentioned that we can't pass select helpers through loop/functions
  - [ ] add vignette about select helpers in general? There are a ton of ways to select or exclude vars and a vignette would be a good place for detailing them.
  - [ ] update NEWS

etiennebacher:
Now works:
``` r
library(datawizard)
i <- "Petal"
foo <- function(data) {
  i <- "Sep"
  find_columns(data, select = starts_with(i))
}
foo(iris)
#> [1] "Sepal.Length" "Sepal.Width"

for (i in "Sepal") {
  head(iris) |>
    find_columns(select = starts_with(i)) |>
    print()
}
#> [1] "Sepal.Length" "Sepal.Width"
```

etiennebacher:
@strengejacke I came back to your implementation and tweaked it to allow using select helpers through functions. Here's a long reprex that summarizes the behavior of `.select_nse`. The PR is not finished (see todo list above).
``` r
library(datawizard)

### Basic behavior (just to show that it keeps working with this PR)
find_columns(iris, select = 1:3)
#> [1] "Sepal.Length" "Sepal.Width"  "Petal.Length"
find_columns(iris, select = c("Sepal.Length", "Petal.Width"))
#> [1] "Sepal.Length" "Petal.Width"
find_columns(iris, select = ~ Sepal.Length + Petal.Width)
#> [1] "Sepal.Length" "Petal.Width"
find_columns(iris, select = starts_with("Sep"))
#> [1] "Sepal.Length" "Sepal.Width"


### Functions

# "i" defined in function environment and not in global environment
foo <- function(data) {
  i <- "Sep"
  find_columns(data, select = starts_with(i))
}
foo(iris)
#> [1] "Sepal.Length" "Sepal.Width"


# "i" defined in function environment *and* in global environment
i <- "Petal"
foo <- function(data) {
  i <- "Sep"
  find_columns(data, select = starts_with(i))
}
foo(iris)
#> [1] "Sepal.Length" "Sepal.Width"


foo <- function(data, arg) {
  find_columns(data, select = arg)
}
foo(iris, starts_with("Sep"))
#> [1] "Sepal.Length" "Sepal.Width"


# "i" defined in global environment but not in function environment
foo <- function(data, arg) {
  find_columns(data, select = arg)
}
foo(iris, starts_with(i))
#> [1] "Petal.Length" "Petal.Width"


### Nested functions

# doesn't work
foo2 <- function(data, arg) {
  foo(data, starts_with(arg))
}
foo2(iris, "Sep")
#> Warning in get0(x, envir = env, inherits = inherits, ifnotfound = myObj):
#> restarting interrupted promise evaluation
#> NULL

# better to do it like that
foo2 <- function(data, arg) {
  foo(data, arg)
}
foo2(iris, starts_with("Sep"))
#> [1] "Sepal.Length" "Sepal.Width"


### Loops

for (i in "Sepal") {
  head(iris) |>
    find_columns(select = starts_with(i)) |>
    print()
}
#> [1] "Sepal.Length" "Sepal.Width"
```

Note that all of this also works when `dplyr` is loaded in the environment so there are no conflicts due to `starts_with` for example. I think it would be useful to write a vignette on all the possibilities to select variables because there are so many. Also, if this PR is merged, we would need to explain how to program with select helpers (basically the examples above).  

Should we run `revdepcheck` on this PR? I fear that potential errors caused by this would be hard to debug.

etiennebacher:
Note to self: wait for #260

IndrajeetPatil:
>  add vignette about select helpers in general? There are a ton of ways to select or exclude vars and a vignette would be a good place for detailing them -> WON'T DO: maybe in another PR

Yes, I'd say this should be done in a separate PR; also to make it easier to review.

etiennebacher:
> Yes, I'd say this should be done in a separate PR; also to make it easier to review.

I already started a vignette but didn't push it in a separate branch for now (just saying so that you don't start a second one)

etiennebacher:
Ok I think this is ready for review. I ran `revdepcheck` locally and everything was fine but I'm not sure I did everything correctly since it was the first time I used it :sweat_smile: Tests should cover most of the cases but it's hard to know with these "environment" features

R/data_find.R
```diff
@@ -64,39 +64,50 @@
...
-#' foo(iris)
 #' ```
 #'
-#' One workaround is to use the `regex` argument, which provides at least a bit
-#' more flexibility than exact matching. `regex` in its basic usage (as seen
-#' below) means that `select` behaves like the `contains("")` select-helper, but
-#' can also make the function more flexible by allowing to define complex
-#' regular expression pattern in `select`.
+#' However, this behavior is limited to a "single-level function". It will not
+#' work in nested functions, like below:
```
IndrajeetPatil, line 93:
```suggestion
#' work in nested functions, like below:

```

R/data_find.R
```diff
@@ -64,39 +64,50 @@
...
+#'   find_columns(data, select = arg)
+#' }
+#' outer <- function(data, arg) {
+#'   inner(data, starts_with(arg))
+#' }
+#' outer(iris, "Sep")
+#' ```
 #'
+#' In this case, it is better to pass the whole select helper as the argument of
+#' `outer()`:
```
IndrajeetPatil, line 105:
```suggestion
#' `outer()`:

```

R/select_helpers.R
```diff
@@ -32,38 +28,46 @@
   }
 
   # check if pattern is a function like "starts_with()"
-  select <- tryCatch(eval(p), error = function(e) NULL)
-  exclude <- tryCatch(eval(p2), error = function(e) NULL)
+  if (!is.null(p) && !insight::safe_deparse(p) %in% ls(globalenv())) {
```
IndrajeetPatil, line 31:
```suggestion
  if (!is.null(p) && !(insight::safe_deparse(p) %in% ls(globalenv()))) {
```

R/select_helpers.R
```diff
@@ -32,38 +28,46 @@
...
+    select <- tryCatch(eval(p), error = function(e) NULL)
+  } else {
+    select <- NULL
+  }
+  if (!is.null(p2) && !insight::safe_deparse(p2) %in% ls(globalenv())) {
```
IndrajeetPatil, line 36: Same as above.

It would actually be better to create a Boolean variable to track this instead of such complex conditional expression.

R/select_helpers.R
```diff
@@ -117,14 +117,84 @@
...
+          }
+        }
+      )
+    }
+  })
+
+  # if element is a select helper but the arg in the select helper is not a
+  # character, e.g starts_with(i), then we need to find the value of this object
+  # by going up all parent.frame() gradually with dynGet()
+  if (!is.null(x) && length(x) == 1 && grepl(.regex_select_helper(), x) && !grepl(paste0(.regex_select_helper(), "\\(\"(.*)\"\\)"), x)) {
```
IndrajeetPatil, line 175:
```suggestion
  if (!is.null(x) && length(x) == 1 && 
  grepl(.regex_select_helper(), x) && 
  !grepl(paste0(.regex_select_helper(), "\\(\"(.*)\"\\)"), x)) {
```

IndrajeetPatil (changes requested):
Amazing work, @etiennebacher!

Just a few nitpicks to address.

R/select_helpers.R
etiennebacher: I can define a boolean before `if` but not sure it's easier to understand. I went with `p_is_expr_not_in_globenv` but not the best name ever

R/select_helpers.R
```diff
@@ -32,38 +28,48 @@
   }
 
   # check if pattern is a function like "starts_with()"
-  select <- tryCatch(eval(p), error = function(e) NULL)
-  exclude <- tryCatch(eval(p2), error = function(e) NULL)
+  p_is_expr_not_in_globenv <- !is.null(p) && !(insight::safe_deparse(p) %in% ls(globalenv()))
+  p2_is_expr_not_in_globenv <- !is.null(p2) && !insight::safe_deparse(p2) %in% ls(globalenv())
```
IndrajeetPatil, line 32:
```suggestion
  is_p_expr_not_in_globenv <- !is.null(p) && !(insight::safe_deparse(p) %in% ls(globalenv()))
  is_p2_expr_not_in_globenv <- !is.null(p2) && !insight::safe_deparse(p2) %in% ls(globalenv())
```

R/select_helpers.R
```diff
@@ -32,38 +28,48 @@
...
+  if (p_is_expr_not_in_globenv) {
```
IndrajeetPatil, line 33:
```suggestion
  if (is_p_expr_not_in_globenv) {
```

R/select_helpers.R
```diff
@@ -32,38 +28,48 @@
...
+    select <- tryCatch(eval(p), error = function(e) NULL)
+  } else {
+    select <- NULL
+  }
+  if (p2_is_expr_not_in_globenv) {
```
IndrajeetPatil, line 38:
```suggestion
  if (is_p2_expr_not_in_globenv) {
```

IndrajeetPatil (approved):
Names for Boolean variables or functions should make clear what true and false means. This can be done using prefixes (is, has, can, etc.).

https://indrajeetpatil.github.io/second-hardest-cs-thing/#39

That's the only change I'd recommend. Otherwise, LGTM.

PR #264, opened 2022-09-13
Selection: using "-" sign with numeric indices removes the columns

### LINKED ISSUE #258: Variables selection: should we remove the use of the minus sign in front of `select`?
Most of the time, the argument “select” can be prefixed with “-” to inverse the selection, e.g:

``` r
find_columns(iris, -c("Sepal.Length"))
```

    ## [1] "Sepal.Width"  "Petal.Length" "Petal.Width"  "Species"

``` r
find_columns(iris, -starts_with("Sep"))
```

    ## [1] "Petal.Length" "Petal.Width"  "Species"

However:

-   this makes the argument `exclude` useless;
-   using “-” in front of the select argument reverses “select” **except** for numeric (prefixing a number with “-” makes the selection start from the end of the dataframe), which is quite misleading IMO:

``` r
find_columns(iris, -1)
```

    ## [1] "Species"

Removing the use of this “-” sign as prefix of select would be a breaking change, but I think it’s important to improve consistency. Another option is to keep the “-” sign, but remove the `exclude` argument and remove the use of -1 to get the last column (perhaps implement something like `last_col()`?).

@easystats/core-team what do you think?

etiennebacher:
Close #258. This PR changes the behavior of select and exclude when there's a minus sign in front of numeric values.

Previously: select the columns starting from the right of the dataset (inconsistent with the behavior of the "-" with other selections).

Now: remove the columns.

``` r
library(datawizard)

find_columns(iris, -c("Sepal.Length", "Petal.Width"))
#> [1] "Sepal.Width"  "Petal.Length" "Species"

find_columns(iris, -1)
#> [1] "Sepal.Width"  "Petal.Length" "Petal.Width"  "Species"

# doesn't work
find_columns(iris, -1:2)
#> Error: You can't mix negative and positive numeric indices in `select` or
#>   `exclude`.

# works
find_columns(iris, -(1:2))
#> [1] "Petal.Length" "Petal.Width"  "Species"

find_columns(iris, -c(1, 3))
#> [1] "Sepal.Width" "Petal.Width" "Species"

find_columns(iris, -starts_with("Sepal", "Petal"))
#> [1] "Species"

find_columns(iris, -ends_with("Length", "Width"))
#> [1] "Species"
```

<sup>Created on 2022-09-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

--- 

Note: it would be better to detect the minus sign at the beginning of `.select_nse()`, make a boolean "invert_selection" and inverse the selection at the complete end of the function. With that, we could remove all the `if (negate)` conditions in the code. 

Problem: it's hard to get the expression of "select", remove the minus sign, and re-assign it to "select" for every type of input. In particular, it's hard to do that with functions with parenthesis (e.g `select = is.numeric()`) because using `substitute` and then `str2lang()` does not convert them back to functions.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 0.5.1.7
+Version: 0.5.1.8
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0003-1995-6531", Twitter = "@patilindrajeets")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,11 @@ BREAKING CHANGES
   
 * `data_merge()` now errors if columns specified in `by` are not in both datasets.
 
+* Using negative values in arguments `select` and `exclude` now removes the columns
+  from the selection/exclusion. The previous behavior was to start the 
+  selection/exclusion from the end of the dataset, which was inconsistent with
+  the use of "-" with other selecting possibilities.
+
 NEW FUNCTIONS
 
 * `data_peek()`, to peek at values and type of variables in a data frame.
--- a/R/select_helpers.R
+++ b/R/select_helpers.R
@@ -92,10 +92,10 @@
   columns <- colnames(data)
 
   # check if negation is requested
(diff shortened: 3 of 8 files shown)


etiennebacher:
The test on ubuntu-latest takes a while to install deps (I don't know why) but I don't see why tests would fail with this one.

IndrajeetPatil (approved):
This is really cool and beautiful code, @etiennebacher!!

Thank you so much for your hard work.

PR #268, opened 2022-09-16
Fix select_helpers with dev version of `tidyselect`

### LINKED ISSUE #267: Test failures with dev tidyselect
I see these failures with dev tidyselect:

```
Failure (test-get_columns.R:273:3): get_columns from other functions
test_fun1(iris, starts_with("Sep")) (`actual`) not equal to iris[c("Sepal.Length", "Sepal.Width")] (`expected`).

`actual` is length 3
`expected` is length 2

`names(actual)`:   "Sepal.Width"  "Petal.Width" "Species"
`names(expected)`: "Sepal.Length" "Sepal.Width"          

`actual$Petal.Width` is a double vector (0.2, 0.2, 0.2, 0.2, 0.2, ...)
`expected$Petal.Width` is absent

`actual$Species` is an S3 object of class <factor>, an integer vector
`expected$Species` is absent

`actual$Sepal.Length` is absent
`expected$Sepal.Length` is a double vector (5.1, 4.9, 4.7, 4.6, 5, ...)

Failure (test-find_columns.R:87:3): find_columns from other functions
test_fun1(iris, starts_with("Sep")) (`actual`) not equal to c("Sepal.Length", "Sepal.Width") (`expected`).

`actual`:   "Sepal.Width"  "Petal.Width" "Species"
`expected`: "Sepal.Length" "Sepal.Width"          
```

Unfortunately I can't reproduce them when running them outside of the full test suite, and the code in `.select_nse()` is rather dense. But I'm happy to help more if you can produce a simple reprex

etiennebacher:
Close #267. 

When `select = starts_with(i)` where `i` is an object in a loop or a custom function, we need to find the value of `i`. Basically, this is done by checking the error message that appears when we try to evaluate `starts_with(i)`. Either the error is something like 'xxx doesn't exist", or it is an error due to `tidyselect` (when a package using it is loaded) because select helpers "have to be used within a *select* environment". 

Apparently there were some slight changes in error messages generated by `tidyselect` dev version, so we didn't catch it anymore.

### DIFF
--- a/R/select_helpers.R
+++ b/R/select_helpers.R
@@ -145,7 +145,7 @@
 
           # if starts_with() et al. come from tidyselect but need to be used in
           # a select environment, then the error doesn't have the same structure.
-          if (is.null(fn) && grepl("tidyselect", e$message)) {
+          if (is.null(fn) && grepl("must be used within a", e$message)) {
             trace <- lapply(e$trace$call, function(x) {
               tmp <- insight::safe_deparse(x)
               if (grepl(paste0("^", .regex_select_helper()), tmp)) {


strengejacke (approved):

strengejacke:
Do we need to add a test for it?

etiennebacher:
There are already tests for select helpers, the problem is that this code relies on the error messages of `tidyselect` so it can break if these messages change. But I don't think we can add more tests for this

PR #269, opened 2022-09-19
Use `insight::format_*` messages

etiennebacher:
I only focused on cases where there were two functions used e.g:
 ```
stop(insight::format_message("error"), call. = FALSE) -> insight::format_error("error")
```
Should we also transition simple `stop()` messages, e.g should we do the following transition?
```
stop("error", call. = FALSE) -> insight::format_error("error")
```


IndrajeetPatil:
Thanks a lot, @etiennebacher!

The HTML-5 check failure is spurious.

PR #273, opened 2022-09-28
Improve data_read()

strengejacke:
- Adds a `convert_factors` argument, to modify behaviour
- Improves stability of correctly labelling factors when converting to factors

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,5 @@
 # datawizard (development version)
 
-* Improved automatic detection of variable type and value labels in `data_read()`.
-
 * `data_read()` gains a `convert_factors` argument, to turn off automatic
   conversion from numeric variables into factors.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,13 +1,13 @@
 # datawizard (development version)
 
-# datawizard 0.6.1
-
-* Updates tests for upcoming changes in the `{tidyselect}` package (#267).
-
 * Improved automatic detection of variable type and value labels in `data_read()`.
 
 * `data_read()` gains a `convert_factors` argument, to turn off automatic
   conversion from numeric variables into factors.
+
+# datawizard 0.6.1
+
+* Updates tests for upcoming changes in the `{tidyselect}` package (#267).
 
 # datawizard 0.6.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@
 * `data_read()` gains a `convert_factors` argument, to turn off automatic
   conversion from numeric variables into factors.
 
+* `data_read()` gains a `skip_empty` argument, skip "empty" (completely missing)
+  variables during import.
+
 # datawizard 0.6.1
 
 * Updates tests for upcoming changes in the `{tidyselect}` package (#267).
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@
 
 * Updates tests for upcoming changes in the `{tidyselect}` package (#267).
 
+* Improved automatic detection of variable type and value labels in `data_read()`.
+
+* `data_read()` gains a `convert_factors` argument, to turn off automatic
+  conversion from numeric variables into factors.
+
 # datawizard 0.6.0
 
 BREAKING CHANGES
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,9 +3,6 @@
 * `data_read()` gains a `convert_factors` argument, to turn off automatic
   conversion from numeric variables into factors.
 
-* `data_read()` gains a `skip_empty` argument, skip "empty" (completely missing)
-  variables during import.
-
 # datawizard 0.6.1
 
 * Updates tests for upcoming changes in the `{tidyselect}` package (#267).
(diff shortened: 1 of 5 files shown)


strengejacke:
Let me get back to this later, I think I can simplify the new code a bit.

R/data_read.R
```diff
@@ -15,6 +15,11 @@
 #' @param path_catalog Character string, path to the catalog file. Only relevant
 #' for SAS data files.
 #' @param encoding The character encoding used for the file. Usually not needed.
+#' @param convert_factors Logical, if `TRUE` (default), numeric variables, where
```
etiennebacher, line 18:
```
if `TRUE` (default), numeric variables where all values have a value label are assumed
```

R/data_read.R
```diff
@@ -50,10 +55,17 @@
...
+#' preserved as `"labels"` attribute. Use `convert_factors = FALSE` to suppress
+#' automatic detection of factors; numeric variables then will not be converted
+#' to factors.
 #'
 #' @export
-data_read <- function(path, path_catalog = NULL, encoding = NULL, verbose = TRUE, ...) {
+data_read <- function(path,
+                      path_catalog = NULL,
+                      encoding = NULL,
+                      convert_factors = TRUE,
```
etiennebacher, line 66: need to add a test for `convert_factors = FALSE`

R/data_read.R
```diff
@@ -106,40 +118,56 @@ data_read <- function(path, path_catalog = NULL, encoding = NULL, verbose = TRUE
...
+              warning = function(w) names(value_labels),
+              error = function(e) names(value_labels)
+            )
+            try(levels(i) <- lvl, silent = TRUE)
+            value_labels <- NULL
+          } else {
+            i <- as.numeric(i)
+          }
+        } else {
+          i <- as.character(i)
```
etiennebacher, line 157: this `else` only applies if `!is.character(i)` is FALSE so I don't think it's needed

NEWS.md
```diff
@@ -4,6 +4,11 @@
 
 * Updates tests for upcoming changes in the `{tidyselect}` package (#267).
 
+* Improved automatic detection of variable type and value labels in `data_read()`.
```
etiennebacher, line 7: `0.6.1` already on cran so need to move these two points to the devel version

R/data_read.R
```diff
@@ -50,10 +55,17 @@
 #' imported value labels will be set as factor levels. Else, if a variable
 #' has _no_ value labels or less value labels than values, the variable is
 #' either converted into numeric or character vector. Value labels are then
-#' preserved as `"labels"` attribute.
+#' preserved as `"labels"` attribute. Use `convert_factors = FALSE` to suppress
+#' automatic detection of factors; numeric variables then will not be converted
+#' to factors.
```
etiennebacher, line 60: I think this section can be simplified:

"`data_read()` is most comparable to `rio::import()`. For data files from SPSS, SAS or Stata, which support labelled data, variables are converted into their most appropriate type. The major difference to `rio::import()` is
that `data_read()` automatically converts fully labelled numeric variables into factors, where imported value labels will be set as factor levels. If a numeric variable has _no_ value labels or less value labels than values, it is not converted to factor. In this case, value labels are preserved as `"labels"` attribute. Character vectors are preserved.  Use `convert_factors = FALSE` to remove the automatic conversion of numeric variables to factors."

strengejacke:
I added another test with a small SPSS sample file, and revisiting the code and looking at the SPSS behaviour, I think that the former code already caught all exceptions and is much simpler. Will revise the code so it only contains the new `convert_factors` argument.

R/data_read.R
```diff
@@ -96,50 +116,76 @@ data_read <- function(path, path_catalog = NULL, encoding = NULL, verbose = TRUE
...
-    stop("The zip-file does not contain any supported file types.", call. = FALSE)
+    insight::format_error("The zip-file does not contain any supported file types.")
   }
 
   path
 }
 
 
+.skip_empty_columns <- function(out, verbose) {
+  # find empty variables
```
etiennebacher, line 127: There is already `datawizard::remove_empty_columns()`.  The difference is that it doesn't remove columns with only `""`, maybe we could modify it instead of making a new helper?

etiennebacher: Actually I'm not sure it's worth adding an argument here when the user could just do `data_read(...) |> remove_empty_columns()`

strengejacke: yeah, true. Forgot we have those functions ;-)

R/data_read.R
```diff
@@ -42,35 +47,51 @@
...
   }
 
   # read data
-  switch(.file_ext(path),
+  out <- switch(.file_ext(path),
     "txt" = ,
     "csv" = .read_text(path, encoding, verbose, ...),
     "xls" = ,
     "xlsx" = .read_excel(path, encoding, verbose, ...),
     "sav" = ,
```
etiennebacher, line 77: @strengejacke can R read `sav` files? If not we can use `haven::read_sav()`

strengejacke: No, we need (and rely on) _haven_ to do this (alternative would be the _foreign_ package). Should also be documented in the help file which packages are used.

etiennebacher: Yes but I don't understand why you specify e.g `.read_spss` for `por` files and you don't specify `.read_sav` for `sav` files

etiennebacher: Ok I think I got it, "sav" also redirects to `.read_spss()`. I'm not used to `switch()`, my bad.

PR #277, opened 2022-09-28
remove dplyr < 0.8 code chunks

strengejacke:
We have had compatibility for `grouped_df()` methods for dplyr < 0.8 - I think we can remove this and only rely on the structure of grouped df's used since dplyr 0.8 and higher.

no tests needed.


etiennebacher:
This should be mentioned in NEWS

IndrajeetPatil:
Thanks, @strengejacke!

PR #290, opened 2022-10-12
Modify attributes replacement for performance gains

etiennebacher:
Some functions should be fast whatever the size of the dataframe, such as `data_relocate()` or `data_reorder()` since it's just about moving columns. However, the way attributes are applied back to the modified dataframe leads to a loss of performance because it also reallocates the row names (see [this SO answer](https://stackoverflow.com/questions/74029805/why-does-adding-attributes-to-a-dataframe-take-longer-with-large-dataframes)). When there are millions of rows, this takes some time. Therefore, we should only reallocate the attributes that are "unusual" for a dataframe. This is done in `.replace_attrs()`.

Current (takes more time for large dataset but it shouldn't):
``` r
library(datawizard)

tmp <- data.frame(x = sample(1:10, size = 2, replace = TRUE),
                   y = sample(1:10, size = 2, replace = TRUE))

tmp2 <- data.frame(x = sample(1:10, size = 1e7, replace = TRUE),
                  y = sample(1:10, size = 1e7, replace = TRUE))

bench::mark(
  datawiz_small = data_relocate(tmp, "x", after = -1),
  datawiz_big = data_relocate(tmp2, "x", after = -1),
  check = FALSE
)
#> # A tibble: 2 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 datawiz_small  300.1µs  390.1µs    2048.    394.1KB    12.5 
#> 2 datawiz_big     57.9ms   64.5ms      15.1    38.1MB     7.54

bench::mark(
  datawiz_small = data_reorder(tmp, "x"),
  datawiz_big = data_reorder(tmp2, "x"),
  check = FALSE
)
#> # A tibble: 2 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 datawiz_small  247.4µs  330.2µs    2229.       12KB    13.0 
#> 2 datawiz_big     50.1ms   59.2ms      15.9    38.1MB     2.66
```

New:
``` r
library(datawizard)

tmp <- data.frame(x = sample(1:10, size = 2, replace = TRUE),
                   y = sample(1:10, size = 2, replace = TRUE))

tmp2 <- data.frame(x = sample(1:10, size = 1e7, replace = TRUE),
                  y = sample(1:10, size = 1e7, replace = TRUE))

bench::mark(
  datawiz_small = data_relocate(tmp, "x", after = -1),
  datawiz_big = data_relocate(tmp2, "x", after = -1),
  check = FALSE
)
#> # A tibble: 2 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 datawiz_small    311µs    380µs     2093.     747KB     14.8
#> 2 datawiz_big      314µs    384µs     2002.      640B     12.7

bench::mark(
  datawiz_small = data_reorder(tmp, "x"),
  datawiz_big = data_reorder(tmp2, "x"),
  check = FALSE
)
#> # A tibble: 2 × 6
#>   expression         min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 datawiz_small    252µs    317µs     2335.    18.6KB     12.6
#> 2 datawiz_big      252µs    328µs     2356.      640B     14.9
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,8 @@ MINOR CHANGES
 
 * When column names are misspelled, most functions now suggest which
   existing columns possibly could be meant.
+  
+* Miscellaneous performance gains.
 
 # datawizard 0.6.2
 
--- a/R/categorize.R
+++ b/R/categorize.R
@@ -319,6 +319,8 @@ categorize.grouped_df <- function(x,
   # works only for dplyr >= 0.8.0
   grps <- attr(x, "groups", exact = TRUE)[[".rows"]]
 
+  attr_data <- attributes(x)
+
   # evaluate arguments
   select <- .select_nse(select,
     x,
@@ -353,7 +355,7 @@ categorize.grouped_df <- function(x,
     )
   }
   # set back class, so data frame still works with dplyr
-  x <- .replace_attrs(x, attributes(x))
+  x <- .replace_attrs(x, attr_data)
   x
 }
 
--- a/R/categorize.R
+++ b/R/categorize.R
@@ -315,7 +315,6 @@ categorize.grouped_df <- function(x,
                                   regex = FALSE,
                                   verbose = TRUE,
                                   ...) {
-  info <- attributes(x)
 
   # works only for dplyr >= 0.8.0
   grps <- attr(x, "groups", exact = TRUE)[[".rows"]]
@@ -354,7 +353,7 @@ categorize.grouped_df <- function(x,
     )
   }
   # set back class, so data frame still works with dplyr
-  attributes(x) <- utils::modifyList(info, attributes(x))
+  x <- .replace_attrs(x, attributes(x))
   x
 }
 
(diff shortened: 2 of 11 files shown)


strengejacke (approved):

IndrajeetPatil:
Thanks, Etienne!

Nothing makes me happier than _O_(1) algorithms 😬

Issue #293, opened 2022-10-14
`data_to_wide` fails with example


etiennebacher:
No problem with current CRAN version (0.6.2) but fails with dev version.

``` r
suppressPackageStartupMessages({
  library(tidyr)
  library(dplyr)
})

weekdays <- c("Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun")

daily <- tibble(
  day = factor(c("Tue", "Thu", "Fri", "Mon"), levels = weekdays),
  value = c(2, 3, 1, 5)
)

daily
#> # A tibble: 4 × 2
#>   day   value
#>   <fct> <dbl>
#> 1 Tue       2
#> 2 Thu       3
#> 3 Fri       1
#> 4 Mon       5

pivot_wider(daily, names_from = day, values_from = value)
#> # A tibble: 1 × 4
#>     Tue   Thu   Fri   Mon
#>   <dbl> <dbl> <dbl> <dbl>
#> 1     2     3     1     5

datawizard::data_to_wide(daily, names_from = "day", values_from = "value")
#> # A tibble: 1 × 4
#>     Tue   Thu   Fri   Mon
#>   <dbl> <dbl> <dbl> <dbl>
#> 1     1     5     3     2
```

etiennebacher:
Caused by #285, not by #292

IndrajeetPatil:
Thanks for finding and fixing the issue, @etiennebacher!

etiennebacher:
Another example failure (works with CRAN version (0.6.2)):
``` r
suppressPackageStartupMessages({
  library(tidyr)
  library(dplyr)
  library(datawizard)
})

weekdays <- c("Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun")

daily <- tibble(
  day = factor(c("Tue", "Thu", "Fri", "Mon"), levels = weekdays),
  value = c(2, 3, 1, 5),
  type = factor(c("A", "B", "B", "A"))
)

daily
#> # A tibble: 4 × 3
#>   day   value type 
#>   <fct> <dbl> <fct>
#> 1 Tue       2 A    
#> 2 Thu       3 B    
#> 3 Fri       1 B    
#> 4 Mon       5 A

pivot_wider(
  daily, 
  names_from = type, 
  values_from = value,
  values_fill = 0
)
#> # A tibble: 4 × 3
#>   day       A     B
#>   <fct> <dbl> <dbl>
#> 1 Tue       2     0
[...]

PR #302, opened 2022-10-22
Fix failing case in `data_to_wide`

etiennebacher:
Related to #293. The rewriting I did in reshaping functions starts to become a mess, I will try to improve it later but at least the example is now fixed.

@IndrajeetPatil if the tests pass, then I'm ok to make the release with the patch for R-devel. I found other situations where the reshaping functions don't match 100% the output of `tidyr` (see #293) but our previous reshaping functions also failed so it's not due to the rewriting. I can take care of these later.

### DIFF
--- a/R/data_to_wide.R
+++ b/R/data_to_wide.R
@@ -165,6 +165,7 @@ data_to_wide <- function(data,
   # create missing combinations
 
   if (not_all_cols_are_selected && incomplete_groups) {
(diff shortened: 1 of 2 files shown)


IndrajeetPatil:
Thanks, @etiennebacher! Yes, we can fix the other pending issues in the next release. No hurry there.

PR #304, opened 2022-10-25
normalize(): do not include bounds for beta regression

### LINKED ISSUE #297: normalize(): do not include bounds for beta regression
easystats is awesome: I was just like "man I do the removal of zeros and ones for beta regressions manually, it would be nice to add that feature to `normalize()`" and it was already there :)

However, I have a question: 

I typically use `data_rescale(x, range = c(0, 1), to = c(0.0001, 0.9999))`, i.e., some liminary bounds based on a low tolerance threshold. I thought one could even use things like `.Machine$double.eps` to really set the zeros and ones to *barely* less than their values, with the assumption that you want a 0 to be as close to 0 as possible in the beta reg model.

In normalize, we say:

>   If `FALSE`, the return value is compressed, using Smithson and Verkuilen's
  (2006) formula `(x * (n - 1) + 0.5) / n`, to avoid zeros and ones in the
  normalized variables. This can be useful in case of beta-regression, where
  the response variable is not allowed to include zeros and ones.

My question was whether this method of compressing was better than the tolerance-based one? And if not, should we include the possibility of another method (e.g., `normalize(x, include_bounds = "tolerance")`)

@mattansb @bwiernik

strengejacke:
Fixes #297

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 0.6.3.1
+Version: 0.6.3.2
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0003-1995-6531", Twitter = "@patilindrajeets")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # datawizard (devel)
 
+MINOR CHANGES
+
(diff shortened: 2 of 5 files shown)


DominiqueMakowski:
What was there to fix?

strengejacke:
> What was there to fix?

The bug I introduced when I replaced `deparse()` with `insight::safe_deparse()` - I also removed the `substitute()`, which I added back then.

strengejacke:
If you refer to the inital post (fixes #297), this is just the default text when you open an PR based on the issue.

DominiqueMakowski (approved):

PR #310, opened 2022-11-06
New functions: `data_unique` & `data_duplicated`

rempsyc:
Follow-up to #301.

---

``` r
devtools::load_all("D:/github/forks/datawizard")
#> ℹ Loading datawizard

df1 <- data.frame(
  id = c(1, 2, 3, 1, 3),
  item1 = c(NA, 1, 1, 2, 3),
  item2 = c(NA, 1, 1, 2, 3),
  item3 = c(NA, 1, 1, 2, 3)
)

data_duplicated(df1, id = "id")
#>   Row id item1 item2 item3 count_na
#> 1   1  1    NA    NA    NA        3
#> 4   4  1     2     2     2        0
#> 3   3  3     1     1     1        0
#> 5   5  3     3     3     3        0

data_unique(df1, id = "id", method = "best")
#> (2 duplicates removed, with method 'best')
#>   id item1 item2 item3
#> 1  1     2     2     2
#> 2  2     1     1     1
#> 3  3     1     1     1

data_unique(df1, id = "id", method = "first")
#> (2 duplicates removed, with method 'first')
#>   id item1 item2 item3
#> 1  1    NA    NA    NA
#> 2  2     1     1     1
#> 3  3     1     1     1

data_unique(df1, id = "id", method = "last")
#> (2 duplicates removed, with method 'last')
#>   id item1 item2 item3
#> 1  1     2     2     2
#> 2  2     1     1     1
#> 3  3     3     3     3
```

Can add tests when approved.

<sup>Created on 2022-11-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # datawizard (devel)
 
+MAJOR CHANGES
+
+* New functions to deal with duplicates: `data_unique` and `data_duplicated`.
```
etiennebacher, line 5: Can you briefly explain the difference between the two?

R/data_duplicated.R
```diff
@@ -0,0 +1,39 @@
...
+#' data_duplicated(df1, id = "id")
+#'
+#' # Filter to exclude duplicates
+#' df2 <- df1[-c(1, 5),]
+#' df2
+
+data_duplicated <- function(data, id) {
+  Row <- seq(nrow(data))
+  data <- cbind(Row, data)
+  dups.index <- data[[id]] %in% data[[id]][duplicated(data[id])]
```
etiennebacher, line 32: This works only if `id` is one variable, right? I have cases where I want to check duplicated observations by country and year, but I guess I can't pass `data_duplicated(data, id = c("country", "year"))`

R/data_unique.R
```diff
@@ -0,0 +1,62 @@
+#' @title Keep only one row from all with duplicated IDs
+#'
+#' @description From all rows with at least one duplicated ID,
+#' keep only one. Methods for selecting the duplicated row are
+#' either the first duplicate, the last duplicate, or the "best"
+#' duplicate (default), based on the duplicate with the fewer
+#' number of NA values. In case of ties, it picks the first
```
etiennebacher, line 7: fewer -> lowest
NA values -> `NA`

R/data_unique.R
```diff
@@ -0,0 +1,62 @@
...
+#' df1 <- data.frame(
+#'    id = c(1, 2, 3, 1, 3),
+#'    item1 = c(NA, 1, 1, 2, 3),
+#'    item2 = c(NA, 1, 1, 2, 3),
+#'    item3 = c(NA, 1, 1, 2, 3)
+#' )
+#'
+#' data_unique(df1, id = "id")
+
+data_unique <- function(data, id, method = "best") {
```
etiennebacher, line 27: same comment as above about `id`

R/data_unique.R
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  good.dups <- data_select(good.dups, og.names)
+  good.data <- data[!duplicated(data[[id]]), ]
+  match.index <- good.data[[id]] %in% good.dups[[id]]
+  good.data[match.index, ] <- good.dups
+
+  dup.msg <- paste0(" duplicates removed, with method '", method, "')")
+  dup.msg <- c(dup.msg, ifelse(dups.n != 69, "", " 69... nice"))
+  message("(", dups.n, dup.msg, sep = "")
```
etiennebacher, line 59: use `insight::format_message()`

etiennebacher:
Thanks @rempsyc ! I made a few comments, the most important one is the fact that `id` cannot take more than one variable currently (unless I'm mistaken, I didn't test) but I think it's important to allow users to pass several vars.

Also, I'd like to do some benchmarks to know if there are parts of the code that need to be improved.

R/data_duplicated.R
```diff
@@ -0,0 +1,39 @@
+#' @title Extract all duplicates
+#'
+#' @description Extract all duplicates, for visual inspection.
+#' Note that it also contains the first occurrence of future
+#' duplicates, unlike [duplicated()] or [dplyr::distinct()]). Also
+#' contains an additional column reporting the number of missing
```
strengejacke, line 6: Can you add backticks to the code functions you're linking here?

R/data_duplicated.R
```diff
@@ -0,0 +1,39 @@
...
+#' )
+#'
+#' data_duplicated(df1, id = "id")
+#'
+#' # Filter to exclude duplicates
+#' df2 <- df1[-c(1, 5),]
+#' df2
+
+data_duplicated <- function(data, id) {
+  Row <- seq(nrow(data))
```
strengejacke, line 30: See lintr

R/data_unique.R
```diff
@@ -0,0 +1,62 @@
...
+  good.data
+
+}
```
strengejacke, line 62: Can you add tests against validated results?

R/data_duplicated.R
rempsyc: You're correct. I had only thought about the scenario of using a single ID variable (thus why I first used `id` instead of `select`). Not sure how I will need to change the code to accommodate that, but `dplyr::distinct` does allow several arguments:

``` r
devtools::load_all("D:/github/forks/datawizard")
#> ℹ Loading datawizard

df1 <- data.frame(
  country = c("US", "CAN", "US", "US", "US"),
  year = c(2022, 2022, 2000, 2022, 2022),
  item1 = c(NA, 1, 1, 2, 3),
  item2 = c(NA, 1, 1, 2, 3),
  item3 = c(NA, 1, 1, 2, 3)
)
df1
#>   country year item1 item2 item3
#> 1      US 2022    NA    NA    NA
#> 2     CAN 2022     1     1     1
#> 3      US 2000     1     1     1
#> 4      US 2022     2     2     2
#> 5      US 2022     3     3     3

dplyr::distinct(df1, country, .keep_all = TRUE)
#>   country year item1 item2 item3
#> 1      US 2022    NA    NA    NA
#> 2     CAN 2022     1     1     1

dplyr::distinct(df1, country, year, .keep_all = TRUE)
#>   country year item1 item2 item3
#> 1      US 2022    NA    NA    NA
#> 2     CAN 2022     1     1     1
#> 3      US 2000     1     1     1

# Doesn't work with data_unique so far
data_unique(df1, id = c("country", "year"), method = "first")
#> Error in .subset2(x, i, exact = exact): subscript out of bounds
```

<sup>Created on 2022-11-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/data_unique.R
rempsyc: That's an easter egg, a [hidden treasure](https://www.dailydot.com/unclick/69-nice-meme-twitter/) of internet culture and one of the most well-known [meme](https://www.urbandictionary.com/define.php?term=69%20nice).

R/data_duplicated.R
etiennebacher: The easiest way is probably to concatenate the id columns together to have a single "temporary_id" column on which you check duplicates, and you remove the "temporary_id" column at the end

R/data_unique.R
etiennebacher: Ah ok, never heard of this one before :smile:

R/data_duplicated.R
rempsyc: I think that's not necessary because roxygen2 [now converts](https://stackoverflow.com/a/59336060/9370662) this notation to code-style writing and also automatically links to the other the functions documentation ([example result](https://rempsyc.remi-theriault.com/reference/extract_duplicates.html)).

R/data_duplicated.R
rempsyc: Thanks. That makes sense. Sometimes with creating temporary columns I'm afraid of overwriting a user column also called something similar like "temporary_id". Is there any sort of convention for this? Of course the weirder the added column name, the less likely it is to overwrite, but still not impossible.

etiennebacher: Right, usually I use "temporary_id" but it's true that it can be an existing name so I should change this habit...

You can create a random name in the function with `new_name <- sample(letters)`, and then use the value of `new_name` as the temporary id.

Here's a way to concatenate columns:
https://github.com/easystats/datawizard/blob/main/R/data_to_wide.R#L148

rempsyc: Also, not sure if it would be worth pulling the `count_na` function out of `data_duplicated`, and if so, whether it should go to `datawizard` or `insight`.

rempsyc: Instead of generating a random name every time the function is run, I think we should agree on a specific temporary column name and use that consistently within `easystats`.

rempsyc: And do you want me to use the argument `select` instead of `id` (like in your original comment in the discussion), or like in the link you provided, `id_cols`? (thanks btw, it's useful)

Same for `method` vs `keep`.

etiennebacher: Yes I think you should use `select`. `id_cols` was just to match the arguments of `tidyr::pivot_` functions. I also prefer `keep`

rempsyc: I now realize that your idea to use a randomly generated name might be better because otherwise one function using the same name will remove that column from the other function using it (as is happening here). However, using a random name creates problems with `data_filter` later on, so not quite sure of the best way forward yet.

rempsyc:
I've added tests and the possibility to select multiple ID variables.

By using `data_select` instead of `data[, select, drop = FALSE]`, I was hoping that we could use the `.select_nse` features without having to explicitly add it to the code. However, it seems like some of them work, but the select-helpers and the unquoted features don't work well (see commented tests).

rempsyc:
Reprex for multiple values:

``` r
devtools::load_all("D:/github/forks/datawizard")
#> ℹ Loading datawizard

df1 <- data.frame(
  country = c("US", "CAN", "US", "US", "US"),
  year = c(2022, 2022, 2000, 2022, 2022),
  item1 = c(NA, 1, 1, 2, 3),
  item2 = c(NA, 1, 1, 2, 3),
  item3 = c(NA, 1, 1, 2, 3)
)
df1
#>   country year item1 item2 item3
#> 1      US 2022    NA    NA    NA
#> 2     CAN 2022     1     1     1
#> 3      US 2000     1     1     1
#> 4      US 2022     2     2     2
#> 5      US 2022     3     3     3

dplyr::distinct(df1, country, year, .keep_all = TRUE)
#>   country year item1 item2 item3
#> 1      US 2022    NA    NA    NA
#> 2     CAN 2022     1     1     1
#> 3      US 2000     1     1     1

data_unique(df1, ~ country + year, "first")
#> (2 duplicates removed, with method 'first')
#>   country year item1 item2 item3
#> 1      US 2022    NA    NA    NA
#> 2     CAN 2022     1     1     1
#> 3      US 2000     1     1     1
```

<sup>Created on 2022-11-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

strengejacke (approved):

etiennebacher:
@rempsyc I fixed the issue with selecting helpers, and I simplified the code for `keep = "best"`

rempsyc:
Awesome, thanks @etiennebacher! 🥳

etiennebacher:
For now, what's missing is the ability to handle grouped data. 

``` r
library(datawizard)

df <- tibble(
  g = c(1, 1, 2, 2),
  x = c(1, 1, 2, 1)
) |> group_by(g)

df |> distinct(x)
#> # A tibble: 3 × 2
#> # Groups:   g [2]
#>       g     x
#>   <dbl> <dbl>
#> 1     1     1
#> 2     2     2
#> 3     2     1

df |> data_unique(x)
#> (2 duplicates removed, with method 'best')
#> # A tibble: 2 × 2
#> # Groups:   g [2]
#>       g     x
#>   <dbl> <dbl>
#> 1     1     1
#> 2     2     2
```

rempsyc:
> For now, what's missing is the ability to handle grouped data.

True. But is it necessary given that including the group variable in select produces an identical result?

``` r
devtools::load_all("D:/github/forks/datawizard")
#> ℹ Loading datawizard

df <- data.frame(
  g = c(1, 1, 2, 2),
  x = c(1, 1, 2, 1)
) |> data_group(g)

df |> dplyr::distinct(x)
#> # A tibble: 3 × 2
#> # Groups:   g [2]
#>       g     x
#>   <dbl> <dbl>
#> 1     1     1
#> 2     2     2
#> 3     2     1

df |> data_unique(c("x", "g"))
#> (1 duplicates removed, with method 'best')
#> # A tibble: 3 × 2
#> # Groups:   g [2]
#>       g     x
#>   <dbl> <dbl>
#> 1     1     1
#> 2     2     2
#> 3     2     1
```

<sup>Created on 2022-11-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

rempsyc:
> There are also some warnings with keep = "last":

In addition to the warnings, the result is incorrect as the last value g for id x 1 should be 2.

rempsyc:
Your example works now:

``` r
devtools::load_all("D:/github/forks/datawizard")
#> ℹ Loading datawizard

df <- data.frame(
  x = c(1, 1, 2, 1),
  g = c(1, 1, 2, 2)
)

df |> data_unique(x, keep = "last")
#> (2 duplicates removed, with method 'last')
#>   x g
#> 1 1 2
#> 3 2 2
```

<sup>Created on 2022-11-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

However, using the `count_na == min(count_na)` method produces this R CMD check NOTE:

```
 Undefined global functions or variables:
    count_na
```

In `dplyr`, one can usually use the `.data` pronoun to get rid of this note without having to rely on `utils::globalVariables`. Is there an official way with `datawizard`?

strengejacke:
> In dplyr, one can usually use the .data pronoun to get rid of this note without having to rely on utils::globalVariables. Is there an official way with datawizard?

No, that's why we usually work with character vectors. :-/

etiennebacher:
> True. But is it necessary given that including the group variable in select produces an identical result?

I would expect `data_unique()` to work in each group separately without having to specify the grouping variable

rempsyc:
Thanks for fixing the note Etienne. Ah, putting the whole expression in quotes, such a good an easy fix, I should have thought of it! 🙈 I really need to get the reflex to use this new feature since it is so awesome.

Regarding adding support for grouped data frames, do you think I should be able to figure it out simply by looking at the [changes you made for `data_filter`](https://github.com/easystats/datawizard/pull/312/files)?

etiennebacher:
Yes it should be similar, I think `data_arrange()` is a better example than `data_filter()` though

etiennebacher:
@rempsyc while I was at it, I added the group behavior and fixed a bug when there are no dups. I think what's left is:
* should we add all args for selection (`exclude`, `regex`, etc.)?
* fix this "temporary_id" name that is not ideal (maybe in another PR because it's not for `data_unique()` only
* some benchmarks to see if we can optimize this (especially with groups)

rempsyc:
Thanks! I was trying my luck on adding the group behaviour but could not quite make it work 😬 And thanks for fixing the bug.

> - should we add all args for selection (exclude, regex, etc.)?

might as well to be consistent with all the other functions
> - fix this "temporary_id" name that is not ideal (maybe in another PR because it's not for data_unique() only

agree for different PR and discuss this with others possibly too
> - some benchmarks to see if we can optimize this (especially with groups)

sounds good!

etiennebacher:
> Should `data_duplicated` also work by group?

Yeah

rempsyc:
Oh no. I pushed the grouped `data_duplicated` support, but it seems like this broke grouped `data_unique` since it uses that internally 🤔

I'll look into it and try to figure it out. And add the other arguments regex, etc.

rempsyc:
Ok! Should be good now. I'll leave the benchmarks to you since I think you're the expert on this! :P

etiennebacher:
Small benchmark:
``` r
suppressPackageStartupMessages({
  library(datawizard)
  library(dplyr)
})

set.seed(123)

test <- tibble(
  id = rep(letters[1:10], each = 100000),
  val1 = sample(c(1:10, NA), 1000000, replace = TRUE),
  val2 = sample(c(letters[1:10], NA), 1000000, replace = TRUE)
) |> 
  group_by(id)

bench::mark(
  datawiz = data_unique(test, "val1", keep = "first"),
  dplyr = dplyr::distinct(test, val1, .keep_all = TRUE),
  check = FALSE,
  iterations = 1
)
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> (99989 duplicates removed, with method 'first')
#> # A tibble: 2 × 6
#> 1 datawiz        8.1s     8.1s     0.123   791.4MB     3.21
#> 2 dplyr        24.1ms   24.1ms    41.4      15.8MB    41.4
```

rempsyc:
So if I understand correctly, `dplyr` way is about 300 hundred times faster? Hum.

etiennebacher:
The objective is not to catch up `dplyr` because they use some C/C++ anyway but I think the speed can be improved

etiennebacher:
The problem was that `data_arrange()` is super slow for ordering large datasets with character vectors and `temporary_id` is a character vector. Calling `data_arrange()` on all `select` columns reduces this problem. 

Now, we have:
``` r
suppressPackageStartupMessages({
  library(datawizard)
  library(dplyr)
})

set.seed(123)

test <- tibble(
  id = rep(letters[1:10], each = 100000),
  val1 = sample(c(1:10, NA), 1000000, replace = TRUE),
  val2 = sample(c(letters[1:10], NA), 1000000, replace = TRUE)
) 

bench::mark(
  datawiz = data_unique(test, "val1", keep = "first"),
  dplyr = dplyr::distinct(test, val1, .keep_all = TRUE),
  iterations = 1
)
#> (999989 duplicates removed, with method 'first')
#> (999989 duplicates removed, with method 'first')
#> # A tibble: 2 × 6
#> 1 datawiz       954ms    954ms      1.05   741.4MB     13.6
#> 2 dplyr          20ms     20ms     50.1     12.1MB      0
```
I'm ok with this, just waiting for github actions

etiennebacher:
Thanks @rempsyc !

PR #329, opened 2022-12-09
Try with the native pipe

IndrajeetPatil:


strengejacke:
I'm not sure if this will work with GHA. If I understood correctly, a _binary_ build with R 4.0 or later will translate the code in the binary file into something w/o the pipe, so it works on older R versions as well. But here, we compile from source with the related R version which is used for testing - so no internal conversion of the pipe is possible for older R versions... ?

PR #337, opened 2022-12-15
Add a roadmap to the website?

etiennebacher:
As discussed in https://github.com/easystats/easystats/issues/330, here's an example of roadmap for `datawizard`. This is just some "placeholder" text so that we can explore if and how we want to add this to the website (how detailed should be the content of this roadmap?, etc.).

Feel free to modify, this is not meant to be merged as-is. Here's what it looks like:

image

image

@rempsyc @IndrajeetPatil WDYT?

### DIFF
--- a/_pkgdown.yaml
+++ b/_pkgdown.yaml
@@ -129,3 +129,9 @@ articles:
     navbar: ~
     contents:
       - standardize_data
+
+navbar:
+  right:
+  - text: "Roadmap"
+    href: articles/roadmap.html
+    icon: fa-solid fa-map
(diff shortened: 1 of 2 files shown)


rempsyc:
Looks good for a start! It is a vignette rather than a plain MD file, which I suppose is more as it should be.

It does look very road-map-like so far and not so to-do-list, but we'll see what Indra envisioned. Perhaps we can add a carefully filtered to-do list under that section for important but more minor elements, if those are the goals.

IndrajeetPatil:
Looks great! Thanks, Etienne. Yeah, this is closer to what I had in mind from a document like this.

etiennebacher:
Not going to finish this

PR #347, opened 2023-01-04
Fix formatting in `data_filter()` doc

### LINKED ISSUE #341: Undo suboptimal formatting for `data_match.R` file
This was deliberate on my part (https://github.com/easystats/datawizard/commit/bb42d9fca319380ec4cdc8701a916c30be4788f8). See my commit message for more details.

etiennebacher:
Close #341


strengejacke:
I personally would prefer to not quote the example and turn styler off in this particular case. As a user, I would be confused and think that I only can use curly braces in quotes if we don't have any other working example.

etiennebacher:
`styler` is disabled in the docs of `data_filter()` now so I'm closing this

PR #367, opened 2023-03-01
Fix converting to factors when more than 9 levels

strengejacke:
When variables are labelled and have more than 9 unqiue values, the former code coerced to character, so sorting of levels was 1, 10, 11, 12, ..., 2, 3, 4, 5, 6, 7, 8, 9. Thus, values no longer matched with value labels.

This PR checks if variable is numeric, and if so, does not coerce to character when converting to factor, so correct order of levels is preserved.

Must create some test data to add a test.

### DIFF
--- a/man/describe_distribution.Rd
+++ b/man/describe_distribution.Rd
@@ -48,11 +48,7 @@ describe_distribution(x, ...)
 \item{x}{A numeric vector, a character vector, a data frame, or a list. See
 \code{Details}.}
 
-\item{...}{Additional arguments to be passed to or from methods.}
-
-\item{centrality}{The point-estimates (centrality indices) to compute.  Character (vector) or list with one or more of these options: \code{"median"}, \code{"mean"}, \code{"MAP"} or \code{"all"}.}
-
-\item{dispersion}{Logical, if \code{TRUE}, computes indices of dispersion related to the estimate(s) (\code{SD} and \code{MAD} for \code{mean} and \code{median}, respectively).}
+\item{...}{Arguments passed down to other functions. Mostly not used yet.}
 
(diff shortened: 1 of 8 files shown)


man/describe_distribution.Rd
```diff
@@ -48,11 +48,7 @@ describe_distribution(x, ...)
 \item{x}{A numeric vector, a character vector, a data frame, or a list. See
 \code{Details}.}
 
-\item{...}{Additional arguments to be passed to or from methods.}
-
-\item{centrality}{The point-estimates (centrality indices) to compute.  Character (vector) or list with one or more of these options: \code{"median"}, \code{"mean"}, \code{"MAP"} or \code{"all"}.}
-
-\item{dispersion}{Logical, if \code{TRUE}, computes indices of dispersion related to the estimate(s) (\code{SD} and \code{MAD} for \code{mean} and \code{median}, respectively).}
```
strengejacke, line 55: Can we please remove this workflow? It ruins working code. @etiennebacher @IndrajeetPatil

IndrajeetPatil: Hmm, I am surprised about this change. It's literally just running `roxygen2::roxygenize()`, so the behaviour on CI/CD should be similar to the local one.

IndrajeetPatil: The workflow is to enforce a consistent style across repos. Just because it made one mistake doesn't mean we should remove it immediately.

IndrajeetPatil: I deliberately ran the style workflow again and I can't reproduce this behaviour. So it was probably a temporary glitch somewhere.

PR #369, opened 2023-03-03
Check for performance gains

### LINKED ISSUE #362: Check for performance gains
Check whether the `unlist()` calls should have `use.names = FALSE` to be faster:
``` r
test <- list(
  hp = rep("hp", 1e6),
  cyl = rep("cyl", 1e6),
  drat = rep("drat", 1e6)
)

bench::mark(
  unlist(test),
  unlist(test, use.names = FALSE),
  check = FALSE,
  iterations = 10
)
#> # A tibble: 2 × 6
#>   expression                           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 unlist(test)                     873.7ms  908.5ms      1.08    76.8MB     2.15
#> 2 unlist(test, use.names = FALSE)   13.4ms   16.3ms     60.8     22.9MB    15.2
```

strengejacke:
Fixes #362

### DIFF
--- a/R/data_codebook.R
+++ b/R/data_codebook.R
@@ -117,7 +117,7 @@ data_codebook <- function(data,
           text_wrap(variable_label, width = variable_label_width),
           "\n",
           fixed = TRUE
-        )))
+        ), use.names = FALSE))
       }
     } else {
       variable_label <- NA
--- a/R/recode_values.R
+++ b/R/recode_values.R
@@ -315,7 +315,10 @@ recode_values.factor <- function(x,
       # pattern: old = new
       # name of list element is old value
 
-      old_values <- paste(deparse(insight::trim_ws(unlist(strsplit(i, ",", fixed = TRUE)))), collapse = ",")
+      old_values <- paste(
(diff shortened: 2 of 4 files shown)


strengejacke:
@etiennebacher I keep track of those functions where I already added `use.names = FALSE`. Not sure about the performance gains, is there any good method to compare this?

etiennebacher:
I usually use `bench::mark()` to see whether a change makes a difference, but here adding `use.names = FALSE` will improve the performance for sure. I think I wanted to see the extent of the perf gain to maybe add a bullet point in NEWS to mention it but I don't think it's actually necessary

strengejacke:
I merged this PR although there may be more instances where we can add `use.names = FALSE`, and that's why I kept the issue open. Feel free to revisit this issue some time in the future, or close it. :-)

>  to maybe add a bullet point in NEWS to mention it but I don't think it's actually necessary

Yeah, I thought about adding a news, too, but I'm not sure this change is "visible" to casual users... I don't mind, feel free to mention this in the news.

PR #371, opened 2023-03-07
New function "recode_into()", similar to "dpyr::case_when()"

strengejacke:


etiennebacher:
For info I can't review before tonight

strengejacke:
> For info I can't review before tonight

No hurry.

strengejacke:
Some examples:

``` r
library(datawizard)
x <- 1:30
recode_into(
  x > 15 ~ "a",
  x > 10 & x <= 15 ~ "b",
  ~ "c"
)
#>  [1] "c" "c" "c" "c" "c" "c" "c" "c" "c" "c" "b" "b" "b" "b" "b" "a" "a" "a" "a"
#> [20] "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a"
# use default argument
recode_into(
  x > 15 ~ "a",
  x > 10 & x <= 15 ~ "b",
  default = "c"
)
#>  [1] "c" "c" "c" "c" "c" "c" "c" "c" "c" "c" "b" "b" "b" "b" "b" "a" "a" "a" "a"
#> [20] "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a"
set.seed(123)
d <- data.frame(
  x = sample(1:5, 30, TRUE),
  y = sample(letters[1:5], 30, TRUE),
  stringsAsFactors = FALSE
)
# from different variables into new vector
recode_into(
  d$x %in% 1:3 & d$y %in% c("a", "b") ~ 1,
  d$x > 3 ~ 2,
  TRUE ~ 0 # the LHS of this formula is arbitrary and can be omitted
)
#>  [1] 1 1 1 0 0 2 2 0 1 1 2 0 0 0 2 1 1 2 1 0 1 1 0 2 0 1 2 2 1 2
# no need to write name of data frame each time
recode_into(
  x %in% 1:3 & y %in% c("a", "b") ~ 1,
  x > 3 ~ 2,
  data = d,
  default = 0
)
#>  [1] 1 1 1 0 0 2 2 0 1 1 2 0 0 0 2 1 1 2 1 0 1 1 0 2 0 1 2 2 1 2
# pipe-style
if (getRversion() >= "4.1.0") {
  d |>
    recode_into(
      data = _,
      x %in% 1:3 & y %in% c("a", "b") ~ 1,
      x > 3 ~ 2,
      default = 0
    )
}
#>  [1] 1 1 1 0 0 2 2 0 1 1 2 0 0 0 2 1 1 2 1 0 1 1 0 2 0 1 2 2 1 2
```

strengejacke:
And seems to work inside functions:

``` r
library(datawizard)

test <- function() {
  set.seed(123)
  d <- data.frame(
    x = sample(1:5, 30, TRUE),
    y = sample(letters[1:5], 30, TRUE),
    stringsAsFactors = FALSE
  )
  recode_into(
    x %in% 1:3 & y %in% c("a", "b") ~ 1,
    x > 3 ~ 2,
    data = d,
    default = 0
  )
}

test()
#>  [1] 1 1 1 0 0 2 2 0 1 1 2 0 0 0 2 1 1 2 1 0 1 1 0 2 0 1 2 2 1 2
```

<sup>Created on 2023-03-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/recode_into.r
```diff
@@ -0,0 +1,109 @@
+#' @title Recode values from one or more variables into a new variable
+#' @name recode_into
+#'
+#' @description
+#' This functions recodes values from one or more variables into a new variable.
+#' It is a convenient function to avoid nested [`ifelse()`] statements.
```
etiennebacher, line 6: We could say it's similar to `dplyr::case_when()`, it'll be easier to understand what it does

R/recode_into.r
```diff
@@ -0,0 +1,109 @@
...
+#' @param data Optional, name of a data frame. This can be used to avoid writing
+#' the data name multiple times in `...`. See 'Examples'.
+#' @param default Indicates the default value that is chosen when no match in
+#' the formulas in `...` is found. If `NULL`, the default value needs to be
+#' determined as last formula in `...`. See 'Examples'.
+#'
+#' @return A vector with recoded values.
+#'
+#' @examples
+#' x <<- 1:30
```
etiennebacher, line 22: Should be `<-`?

etiennebacher (changes requested):

tests/testthat/test-recode_into.R
```diff
@@ -0,0 +1,45 @@
...
+  expect_identical(out1, out2)
+  d <<- mtcars
+  out1 <- recode_into(
+    mpg > 20 & cyl == 6 ~ 1,
+    mpg <= 20 ~ 2,
+    default = 0,
+    data = d
+  )
+  expect_identical(out1, out2)
+})
```
etiennebacher, line 45: Also need to add a test that it works inside function

R/recode_into.r
strengejacke: No, else examples are failing. It's again one of these environment issues.

R/recode_into.r
```diff
@@ -0,0 +1,109 @@
...
+  # iterate all expressions
+  for (i in seq_len(n_params)) {
+    if (is.null(data)) {
+      index <- eval(dots[[i]][[2]])
+      value <- eval(dots[[i]][[3]])
+    } else {
+      index <- with(data, eval(dots[[i]][[2]]))
+      value <- with(data, eval(dots[[i]][[3]]))
+    }
+    out[index] <- value
```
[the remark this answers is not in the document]
strengejacke: Should we warn the user, or error?

R/recode_into.r
etiennebacher: Ah ok

R/recode_into.r
```diff
@@ -0,0 +1,116 @@
...
+      all_recodes <- typeof(eval(dots[[i]][[3]]))
+    } else {
+      all_recodes <- typeof(with(data, eval(dots[[i]][[3]])))
+    }
+    # if we have mixed types, warn user
+    if (!(all_recodes == all_recodes[1]) && verbose) {
+      insight::format_warning(
+        "Not all recode values are of the same type.",
+        "Trying to coerce all values into a common type, please check your recoded variable!"
+      )
```
etiennebacher, line 99: I think this should be an error, I don't see a single case where this would be expected by the user

tests/testthat/test-recode_into.R
strengejacke: Let's see whether we get scoping problems...

etiennebacher: there's a mistake in the expected output

R/recode_into.r
etiennebacher: need to fix the message as well since we no longer try to coerce

tests/testthat/test-recode_into.R
strengejacke: yeah, I copied the code from the mtcars-results

tests/testthat/test-recode_into.R
```diff
@@ -0,0 +1,81 @@
...
+  )
+  expect_identical(out, c("c", "c", "b", "b", "b", "a", "a", "a", "a", "a"))
+})
+
+test_that("recode_into, check mixed types", {
+  x <<- 1:10
+  expect_error(out <- recode_into( # nolint
+    x > 5 ~ 1,
+    x > 2 & x <= 5 ~ "b"
+  ))
```
etiennebacher, line 16: Last thing for me, can you add the regexp for `expect_error` / `_warning` / `_message`?

After that and once the examples are fixed I'm ok to merge

etiennebacher:
The placeholder `_` appeared in 4.2.0, not 4.1.0

strengejacke:
ok, fixed

etiennebacher:
I don't get why examples ignore the `if (getRversion()`, maybe we can just remove the pipe and use the data directly in `recode_into()`

strengejacke:
> I don't get why examples ignore the `if (getRversion()`, maybe we can just remove the pipe and use the data directly in `recode_into()`

Can't follow, where is it ignored? In pkgdown? Else, we can remove that particular example, we already have a data frame example included.

strengejacke:
I think all tests are passing?

etiennebacher:
> Can't follow, where is it ignored? In pkgdown? Else, we can remove that particular example, we already have a data frame example included.

Nope in R CMD check with oldrel, see [here](https://github.com/easystats/datawizard/actions/runs/4357637293/jobs/7617191548)

> I think all tests are passing?

See link

strengejacke:
Ah, ok. Wasn't looking at oldrel-checks, as these fail anyway ;-)

etiennebacher (approved):
Failures are unrelated

etiennebacher:
Thank you @strengejacke !

PR #372, opened 2023-03-07
New "add_labs()" function, to easily add variable and value labels as attributes

strengejacke:
After recoding etc. it is sometime necessary to add new value and variable labels. You can do this with other packages, but why not a little helper in _datawizard_?


R/add_labels.R
```diff
@@ -0,0 +1,157 @@
+#' @title Add variable and value labels
+#' @name add_labels
```
etiennebacher, line 2: I know you wrote it but might be good to avoid name conflicts with `sjlabelled::add_labels()`?

R/add_labels.R
```diff
@@ -0,0 +1,157 @@
...
+#' )
+#'
+#' # labelling using named vectors
+#' data(iris)
+#' out <- add_labels(
+#'   iris$Species,
+#'   variable = "Labelled Species",
+#'   values = c(`setosa` = "Spec1", `versicolor` = "Spec2", `virginica` = "Spec3")
+#' )
+#' str(out)
```
etiennebacher, line 39: The names and values of `attr("labels")` are inverted compared to `sjlabelled::add_labels()`:
``` r
df <- iris[c(1, 55, 108), ]

datawizard::add_labels(
  df$Species,
  variable = "Labelled Species",
  values = c(`setosa` = "Spec1", `versicolor` = "Spec2", `virginica` = "Spec3")
)
#> [1] setosa     versicolor virginica 
#> attr(,"label")
#> [1] Labelled Species
#> attr(,"labels")
#>      Spec1      Spec2      Spec3 
#>     setosa versicolor  virginica 
#> Levels: setosa versicolor virginica

sjlabelled::add_labels(
  df$Species,
  labels = c(`setosa` = "Spec1", `versicolor` = "Spec2", `virginica` = "Spec3")
)
#> Warning in order(as.numeric(all.labels)): NAs introduced by coercion
#> [1] setosa     versicolor virginica 
#> attr(,"labels")
#>     setosa versicolor  virginica 
#>      Spec1      Spec2      Spec3 
#> Levels: setosa versicolor virginica
```

R/add_labels.R
```diff
@@ -0,0 +1,157 @@
...
+#'
+#' # Partial labelling
+#' x <- 1:5
+#' add_labels(
+#'   x,
+#'   variable = "My x",
+#'   values = c(`1` = "lowest", `5` = "highest"),
+#'   verbose = FALSE
```
etiennebacher, line 47: `verbose = FALSE` doesn't change anything here?

R/add_labels.R
```diff
@@ -0,0 +1,157 @@
...
+  # do we have a names vector for "values"?
+  # else check if number of labels and values match
+  if (is.null(names(values))) {
+    if (length(values) == length(unique_values)) {
+      labels <- stats::setNames(unique_values, values)
+    } else if (verbose) {
+      insight::format_alert(
+        "Cannot add labels. Number of unique values and number of value labels are not equal.",
+        sprintf("There are %i unique values and %i provided labels.", length(unique_values), length(values))
+      )
```
etiennebacher, line 89: should it be an error instead?

R/add_labels.R
```diff
@@ -0,0 +1,157 @@
...
+    }
+  } else {
+    # check whether we have matches of labels and values
+    matching_labels <- names(values) %in% unique_values
```
etiennebacher, line 93: perhaps throw an error if all values are not named, it would be clearer than this:
```r
> add_labels(c(1, 2), values = c("foo", `1` = "lowest"))
Following labels were associated with values that don't exist:
  `foo ()`
[1] 1 2
attr(,"labels")
lowest 
     1 
```

R/add_labels.R
```diff
@@ -0,0 +1,157 @@
...
+#'
+#' # Partial labelling
+#' x <- 1:5
+#' add_labels(
+#'   x,
+#'   variable = "My x",
+#'   values = c(`1` = "lowest", `5` = "highest"),
+#'   verbose = FALSE
+#' )
+#' @export
```
etiennebacher, line 49: I think it'd be useful to have an example with a dataframe

R/add_labels.R
```diff
@@ -0,0 +1,157 @@
...
+add_labels.character <- add_labels.numeric
+
+#' @rdname add_labels
+#' @export
+add_labels.data.frame <- function(x,
+                                  select = NULL,
+                                  exclude = NULL,
+                                  values = NULL,
+                                  ignore_case = FALSE,
+                                  append = FALSE,
```
etiennebacher, line 125: what is `append`? it's not in the docs

tests/testthat/test-add_labels.R
```diff
@@ -0,0 +1,48 @@
...
+    verbose = FALSE
+  )
+  expect_identical(attributes(out)$label, "My x")
+  expect_identical(attributes(out)$labels, c(lowest = 1, highest = 5))
+})
+
+test_that("add_labels, messages", {
+  x <- 1:5
+  expect_message(add_labels(x, values = c(`1` = "lowest", `6` = "highest")))
+  expect_message(add_labels(x, variable = 1, values = c(`1` = "lowest", `6` = "highest")))
```
etiennebacher, line 40: I don't understand why but the second `expect_message()` prints the message in `testthat` output

etiennebacher (changes requested):

R/add_labels.R
strengejacke: No, it's handled internally, and I wanted to follow the pattern `"value = label"`.  in `sjlabelled::add_labels()`, it is `"label = value"`. What do you think makes more sense? The implementation here follows the data structure from `haven::labelled()`.

R/add_labels.R
strengejacke: haha, I was only thinking of `set_label()` and `set_labels()`. Damn... Any suggestions for a new name?

tests/testthat/test-add_labels.R
strengejacke: I guess because we have two messages here, one for the wrong type in `variable`, one for non-matching value labels. I'll check after I addressed your comments, in case we change msg to error.

R/add_labels.R
strengejacke: Would work like in other recoding/transformation functions, but maybe is less useful here? I remove it for now.

R/add_labels.R
etiennebacher: `append_label`? `put_label()`? I'm not convinced by either

R/add_labels.R
etiennebacher: I meant in the output, the names and the values of the "labels" vector are inverted.

In the function call, I personally prefer `"value = label"` but maybe it's better to keep the order in which `sjlabelled` does it to reduce the cognitive load

strengejacke: They're inverted because the correct call to that function would be

```r
sjlabelled::add_labels(
  df$Species,
  labels = c("Spec1" = "setosa", "Spec2" = "versicolor", "Spec3" = "virginica")
)
```

> to reduce the cognitive load

We already have this in `recode_values()`, which has the opposite pattern as compared to `sjmisc::rec()`. I'd say we can have a "breaking change" here.

R/add_labels.R
etiennebacher: also `assign_label()`, `apply_label()`, `give_label()`?

strengejacke: chat GPT suggests `map_labels()`.

strengejacke: or something like `data_labels()` or similar...?

etiennebacher: I think we keep `data_` functions for dataframes only (and not a big fan of `map_labels`)

strengejacke: or we don't care if we add another name conflict? There are already some with _sjlabelled_ and _sjmisc_. The long term goal would be to soft-deprecate sjverse and switch to easystats, anyway ;-)

R/add_labels.R
etiennebacher: Ah ok, it's clearer now, thanks. 

True and `dplyr::recode()` also does `old = new`. What is best to keep consistency both inside `datawizard` and with other packages? 

Speaking of `recode_values()`, I have 2 questions: 1) why do we need a list in the second arg while `recode_into()` works fine with the dots? 2) do we need both `recode_values()` and `recode_into()`?

R/add_labels.R
etiennebacher: or just `add_lab()` and `add_labs()` ?

R/add_labels.R
strengejacke: First, `recode_values()` is a generic, where dots are somewhat "reserved". Second, not necessarily. `recode_values()` is a bit shorter and maybe more "intuitive" or clearer for the case of recoding just one variable. But yes, I think you can achieve all recodings you could do in `recode_values()` with `recode_into()`, as well.

R/add_labels.R
etiennebacher: yeah but there's `ggplot2::labs()` so it doesn't come out of nowhere ^^

strengejacke: > or just add_lab() and add_labs()

I would prefer to have just one function that sets value and/or variable labels depending on the provided arguments. So it would be `add_labs()`.

man/describe_distribution.Rd
```diff
@@ -71,8 +67,6 @@ the first centrality index (which is typically the median).}
 \item{iterations}{The number of bootstrap replicates for computing confidence
 intervals. Only applies when \code{ci} is not \code{NULL}.}
 
-\item{threshold}{For \code{centrality = "trimmed"} (i.e. trimmed mean), indicates the fraction (0 to 0.5) of observations to be trimmed from each end of the vector before the mean is computed.}
```
strengejacke: @IndrajeetPatil This workflow again introduced errors automatically. :-/

strengejacke:
@etiennebacher ready to re-review, I think.

man/describe_distribution.Rd
IndrajeetPatil: Hmm, I don't get why this workflow misbehaves only with this file in the entire ecosystem. 
There must be something different about it.

IndrajeetPatil: But, until we figure that out, we can delete the workflow file.

R/add_labs.R
```diff
@@ -0,0 +1,171 @@
...
+
+#' @rdname add_labs
+#' @export
+add_labs.numeric <- function(x, variable = NULL, values = NULL, verbose = TRUE, ...) {
+  # add variable label
+  if (!is.null(variable)) {
+    if (is.character(variable) && length(variable) == 1) {
+      attr(x, "label") <- variable
+    } else if (verbose) {
+      insight::format_alert(
```
etiennebacher, line 81: throw an error rather than a message

tests/testthat/test-add_labs.R
```diff
@@ -0,0 +1,49 @@
...
+  expect_error(expect_message(add_labs(x, variable = 1, values = c(`1` = "lowest", `6` = "highest"))))
+  expect_error(add_labs(x, values = c("a", "b", "c")))
+  expect_message(add_labs(x, variable = 1, values = c(`1` = "lowest", `5` = "highest")))
+})
+
+test_that("add_labs, data frame", {
+  data(iris)
+  out <- add_labs(iris, "Species", values = c("a", "b", "c"))
+  expect_identical(attributes(out$Species)$labels, c(a = "setosa", b = "versicolor", c = "virginica"))
+})
```
etiennebacher, line 49: can you add a test with several vars in `select`?

tests/testthat/test-add_labs.R
```diff
@@ -0,0 +1,49 @@
...
+    verbose = FALSE
+  )
+  expect_identical(attributes(out)$label, "My x")
+  expect_identical(attributes(out)$labels, c(lowest = 1, highest = 5))
+})
+
+test_that("add_labs, error and messages", {
+  x <- 1:5
+  expect_error(add_labs(x, values = c(`1` = "lowest", `6` = "highest")))
+  expect_error(expect_message(add_labs(x, variable = 1, values = c(`1` = "lowest", `6` = "highest"))))
```
etiennebacher, line 40: need to update this test if we throw an error for incorrect `variable`

etiennebacher (approved):
Looks good to me, you can merge once the comments are addressed

R/add_labs.R
etiennebacher: and erroring should not depend on `verbose`

R/add_labels.R
IndrajeetPatil: Of the mentioned suggestions, I like `set_labels()` the best.

strengejacke: That would also conflict with _sjlabelled_.

IndrajeetPatil: Damn.
How different are these functions from _sjlabelled_? 
Is there a possibility to re-export the _datawizard_ versions of these functions from _sjlabelled_?

PR #376, opened 2023-03-13
Add missing select helpers

### LINKED ISSUE #133: Missing select helpers
See #128 

Needs select helpers:
- [x] https://easystats.github.io/datawizard/reference/convert_data_to_numeric.html
- [ ] ~~https://easystats.github.io/datawizard/reference/demean.html~~ (won't do, see #376)
- [ ] https://easystats.github.io/datawizard/reference/winsorize.html
- [x] https://easystats.github.io/datawizard/reference/describe_distribution.html
- [ ] https://easystats.github.io/datawizard/reference/replace_nan_inf.html

Needs to move secondary select-args (basically, data frame method should have argument order of numeric/vector, followed by select/exclude):
- [x] https://easystats.github.io/datawizard/reference/normalize.html
- [x] https://easystats.github.io/datawizard/reference/ranktransform.html

For some functions, select/exclude come after some main arguments, but not at the end. I'm not sure if this is a problem?

etiennebacher:
Close #133 

Still need to do `demean()` but I think it should be renamed `data_demean()` since it only accepts dataframes


strengejacke:
> > Still need to do demean() but I think it should be renamed data_demean() since it only accepts dataframes
> 
> @IndrajeetPatil @strengejacke do you agree on this?

It's a decision between "accepts only data frames" and "belongs to the center and standardize family". You could argue therefore to keep this name.

etiennebacher:
Whether we rename it or not, introducing select helpers in `demean()` would lead to breaking changes because we allow interaction between selected variables, which is not supported by `.select_nse()` at the moment (and maybe never will?):
``` r
library(datawizard)

dat <- data.frame(
  a = c(1, 2, 3, 4),
  x = c(4, 3, 3, 4),
  y = c(1, 2, 1, 2),
  ID = c(1, 2, 3, 1)
)
demean(dat, select = c("a", "x*y"), group = "ID")
#>   a_between x_y_between a_within x_y_within
#> 1       2.5           6     -1.5         -2
#> 2       2.0           6      0.0          0
#> 3       3.0           3      0.0          0
#> 4       2.5           6      1.5          2
```

strengejacke:
I think we can just leave `demean()` as is. select-helpers are probably not that necessary here, because you usually have a limited set of variables in your model and even less that you want to demean (in practice). I think demeaning interaction terms (i.e. allow interactions in `select`) is here more important that the select-helper feature.

etiennebacher:
Ok, `demean()` won't change and `winsorize()` is still being worked on in other PRs so I don't modify it here.

PR #383, opened 2023-03-17
Add error if `unnormalize` not supported

etiennebacher:
There were two `unnormalize.numeric()`, I removed the wrong one and added a `.default` method that errors.

@strengejacke I see that you prefer putting messages in this case (e.g `reverse()`) but I simply think that if a function can't work with an input, it should error


IndrajeetPatil (approved):
LGTM. Thanks!

strengejacke:
> @strengejacke I see that you prefer putting messages

Yes, sometime, when you have a data frame input, some columns maybe don't work. In that case I prefer to "skip" them instead of getting an error, it's just a convenient thing.

PR #386, opened 2023-03-22
New function to convert value labels into factor levels

strengejacke:
``` r
library(datawizard)
# use labels as levels
data(efc)
# create factor
x <- as.factor(efc$c172code)
# add value labels - these are not factor levels yet
x <- add_labs(x, values = c(`1` = "low", `2` = "mid", `3` = "high"))
levels(x)
#> [1] "1" "2" "3"
data_tabulate(x)
#> x <categorical>
#> # total N=100 valid N=90
#> 
#> Value |  N | Raw % | Valid % | Cumulative %
#> ------+----+-------+---------+-------------
#> 1     |  8 |  8.00 |    8.89 |         8.89
#> 2     | 66 | 66.00 |   73.33 |        82.22
#> 3     | 16 | 16.00 |   17.78 |       100.00
#> <NA>  | 10 | 10.00 |    <NA> |         <NA>

x <- labs_to_levels(x)
levels(x)
#> [1] "low"  "mid"  "high"
data_tabulate(x)
#> x <categorical>
#> # total N=100 valid N=90
#> 
#> Value |  N | Raw % | Valid % | Cumulative %
#> ------+----+-------+---------+-------------
#> low   |  8 |  8.00 |    8.89 |         8.89
#> mid   | 66 | 66.00 |   73.33 |        82.22
#> high  | 16 | 16.00 |   17.78 |       100.00
#> <NA>  | 10 | 10.00 |    <NA> |         <NA>
```

<sup>Created on 2023-03-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


strengejacke:
@etiennebacher I'm not sure whether we should just add an argument to `to_factor.factor()`, or have a new function for this feature. Currently, `to_factor()` uses value labels as factor levels, except for factors, which are ignored. We could have a new argument that explicitly sets value labels as factor levels for factors in `to_factor()`, too, or use this new function.

strengejacke:
I have moved the main code to `to_factors()`, where we have the helper `.value_labels_to_levels()` now. If we decide to add this feature via argument `to_factor()`, it can be easily integrated into `to_factor.factor()`.

R/labs_to_levels.R
```diff
@@ -0,0 +1,116 @@
+#' @title Convert value labels into factor levels
+#' @name labs_to_levels
+#'
+#' @details
+#' Convert variables or data into factors and uses value labels as factor levels.
+#' This function is similar to `to_factor()`, however, it is specially designed
+#' for factors with value labels attributes. While `to_factor()` returns factors
+#' as is, `labs_to_levels()` always uses value labels as factor levels, even
```
etiennebacher, line 8: "specially designed for factors with value labels attributes" -> contradicts the NEWS bullet that says it converts data into factors

R/labs_to_levels.R
```diff
@@ -0,0 +1,116 @@
...
+#' for factors.
+#'
+#' @param x A data frame or vector.
+#' @param remove_attr Logical, if `"labels"` attribute should be removed after
+#' value labels are set as factor levels.
+#' @param ... Arguments passed to or from other methods, in this in particular
+#' to `to_factor()`.
```
etiennebacher, line 15: missing word

R/labs_to_levels.R
```diff
@@ -0,0 +1,116 @@
...
+#' @inheritParams find_columns
+#' @inheritParams categorize
+#'
+#' @return A factor, or a data frame of factors.
+#'
+#' @examples
+#' str(labs_to_levels(iris))
+#'
```
etiennebacher, line 23: This example is weird to me and actually I think `labs_to_levels()` does too much. It doesn't *only* converts labels to levels, it also converts to factor (which makes sense since we're talking about factors) but I think a function should do one thing only.

If someone wants to convert to factor, they should use `to_factor()`, potentially with the option to convert labs to levels.

This is what I have in mind:

* I have a labelled factor but labels are not levels -> `labs_to_levels()`
* I have an unlabelled factor -> `labs_to_levels()` errors 
* for other things, `to_factor(labs_to_levels = TRUE)` or `to_factor(labs_to_levels = FALSE)` + `labs_to_levels()`

R/labs_to_levels.R
```diff
@@ -0,0 +1,116 @@
...
+#' levels(x)
+#' data_tabulate(x)
+#'
+#' x <- labs_to_levels(x)
+#' levels(x)
+#' data_tabulate(x)
+#' @export
+labs_to_levels <- function(x, ...) {
+  UseMethod("labs_to_levels")
+}
```
etiennebacher, line 39: For all methods, I think it should error if there are no labels

R/to_factor.R
```diff
@@ -49,22 +49,15 @@ to_factor.numeric <- function(x, ...) {
...
-  }
-
-  # add back variable label
+  # add back labels
   attr(x, "label") <- variable_label
-  x
+  attr(x, "labels") <- value_labels
+
+  # value labels to factor levels
+  .value_labels_to_levels(x)
```
etiennebacher, line 60: this should be an option for the user? (could be the default though)

R/to_factor.R
```diff
@@ -125,3 +118,34 @@ to_factor.data.frame <- function(x,
...
+  # check positions of matching values and levels
+  levels_in_labs <- stats::na.omit(match(value_labels, levels(x)))
+  labs_in_levels <- stats::na.omit(match(levels(x), value_labels))
+  # sanity check - if labelled values and levels don't match
+  if (!length(levels_in_labs) || !length(labs_in_levels)) {
+    if (verbose) {
+      insight::format_alert(
+        "Could not use value labels as factor levels.",
+        "Labelled values and factor levels didn't macth."
+      )
```
etiennebacher, line 142: "Labelled values and factor levels didn't macth." -> "Labelled values and factor levels had no match."

Also "macth" -> "match"

R/to_factor.R
```diff
@@ -125,3 +118,34 @@ to_factor.data.frame <- function(x,
...
+    }
+    return(x)
```
etiennebacher, line 144: should have a message if not all levels had a matching label?

etiennebacher (changes requested):

R/labs_to_levels.R
strengejacke: Should we then better add this as an option to `to_factor()`, instead of having two functions? Or would it be weird that `to_factor()` actually changes factors when argument `labs_to_levels = TRUE`?

R/to_factor.R
strengejacke: Yes, that would be the alternative to having a completely new function.

R/labs_to_levels.R
```diff
@@ -0,0 +1,116 @@
...
+#' # add value labels - these are not factor levels yet
+#' x <- add_labs(x, values = c(`1` = "low", `2` = "mid", `3` = "high"))
+#' levels(x)
+#' data_tabulate(x)
+#'
+#' x <- labs_to_levels(x)
+#' levels(x)
+#' data_tabulate(x)
+#' @export
+labs_to_levels <- function(x, ...) {
```
IndrajeetPatil, line 37: Let's not abbreviate. `labels_to_levels()` is a more informative name. Saving two characters is not worth the user confusion.

IndrajeetPatil: These are `labs` for me 😅 

image

strengejacke: You can discuss this with @etiennebacher ;-)

R/labs_to_levels.R
etiennebacher: but then in the first case (labelled factor but labels are not levels), you would need to pass a factor into `to_factor()` so that would be weird. I think this `labs_to_levels()` (or whatever the name is) can exist but its goal should be limited to only put labels as levels

R/labs_to_levels.R
etiennebacher: > Sorry, I missed this. But I'd change this to add_labels() as well.

So what about name conflict with `sjlabelled::add_labels()`? There are a bunch of propositions in the linked comment if you wanna see alternatives

R/labs_to_levels.R
strengejacke: I personally am not a big fan of error, but rather would use warnings. In our field, we often have data sets, which are not very clean, so some variables *may* have value labels while others do not. And if you want to iterate over a complete (large) data frame, it can be pain picking out and excluding those variables that don't work. These are rather rare occasions - but *if* it happens, users are mostly unable to deal with such errors (I teach R to absolute newcomers...)

strengejacke:
I'll add tests once we decided on the function and argument name.

R/labs_to_levels.R
IndrajeetPatil: Of the remaining suggestions, I like apply_labels() and assign_labels()

strengejacke: Yes, both sound good to me

etiennebacher: Honestly I find both less natural than `add_labs()` but if I had to choose I'd go with `assign_labels()`. What about reexporting the function from `sjlabelled` as @IndrajeetPatil proposed? Is this an option?

strengejacke: But then we would need to import _sjlabelled_, or am I missing something?

IndrajeetPatil: I mean the other way round: you re-export `datawizard::add_labels()` as `sjlabelled::add_labels()`. 
But the hidden assumption here is that their behaviour are comparable.

strengejacke: No. `datawizard::add_labels()` is more like `sjlabelled::set_label()` and `sjlabelled::set_labels()` (i.e. it covers both cases).

`sjlabelled::add_labels()` does not drop existing value labels, so you can overwrite and add labels to an already labelled variable:

> add_labels() adds labels to the existing value labels of x, however, unlike [set_labels](https://strengejacke.github.io/sjlabelled/reference/set_labels.html), it does not remove labels that were not specified in labels. add_labels() also replaces existing value labels, but preserves the remaining labels.

strengejacke: >but if I had to choose I'd go with assign_labels()

>  Let's go with apply_labels()?

😄

IndrajeetPatil: I change my mind every day when the perfect name is not available 😭

strengejacke: Don't worry, the first vote was from @etiennebacher, the second yours. So you both just have to come to an agreement.

strengejacke:
This PR is ready to re-review.

tests/testthat/test-labels_to_levels.R
```diff
@@ -0,0 +1,36 @@
+test_that("labels_to_levels, numeric", {
+  expect_message(
+    labels_to_levels(1:10),
+    regex = "only works"
+  )
+})
+
+test_that("labels_to_levels, factor", {
```
IndrajeetPatil, line 8: Should there also be a test for ordered factors?

NEWS.md
```diff
@@ -4,6 +4,11 @@ BUG FIXES
 
 * Fix test for CRAN check on Mac OS arm64
 
+NEW FUNCTIONS
+
+* `labels_to_levels()`, to convert levels of factors (or factors in data frames)
+  by setting value labels as factor levels.
+
```
etiennebacher, line 11: I think this is clearer: "to use value labels of factors as their levels."

R/labels_to_levels.R
```diff
@@ -0,0 +1,105 @@
+#' @title Convert value labels into factor levels
+#' @name labels_to_levels
+#'
+#' @details
+#' Changes levels of factors (or factors in data frames) by setting value labels
+#' as factor levels.
```
etiennebacher, line 6: same as point in NEWS

tests/testthat/test-labels_to_levels.R
```diff
@@ -0,0 +1,42 @@
...
+  data(efc)
+  x <- as.factor(efc$c172code)
+  attr(x, "labels") <- c("low" = 1, "mid" = 2, "high" = 3)
+  x <- datawizard::labels_to_levels(x)
```
etiennebacher, line 12: no need to attach datawizard here and below

etiennebacher (approved):
Minor things, and can you also bump the version?

Otherwise, looks great, thanks @strengejacke

PR #387, opened 2023-03-22
make sure we really import a data frame

strengejacke:
`.rds` files, which `data_read()` passes to `rio::import()`, can be of arbitrary type. We need to check if we actually imported a data frame before we process the data.


tests/testthat/test-data_read.R
```diff
@@ -171,7 +171,7 @@ test_that("data_read", {
     }
   )), regex = "0 out of 5")
 
-  expect_s3_class("data.frame")
+  expect_s3_class(d, "data.frame")
   expect_identical(dim(d), as.integer(c(2, 5)))
```
IndrajeetPatil, line 175:
```suggestion
  expect_identical(dim(d), c(2L, 5L))
```

R/data_read.R
```diff
@@ -158,6 +158,23 @@ data_read <- function(path,
 # process imported data from SPSS, SAS or Stata -----------------------
 
 .post_process_imported_data <- function(x, convert_factors, verbose) {
+  # for "unknown" data formats (like .RDS), which still can be imported via
```
etiennebacher, line 161: shouldn't this be directly  in `.read_unknown()` since it's the only case where the import can be other than a dataframe? 

If we import an SAS or SPSS file and don't have a dataframe, we should error and not warn

R/data_read.R
```diff
@@ -158,6 +158,23 @@ data_read <- function(path,
...
+  # "rio::import()", we must check whether we actually have a data frame or
+  # not. Else, tell user.
+  if (!is.data.frame(x)) {
+    tmp <- tryCatch(as.data.frame(x, stringsAsFactors = FALSE), error = function(e) NULL)
```
etiennebacher, line 165: use `.coerce_to_dataframe()` instead

etiennebacher (approved):
Some comments, otherwise LGTM

R/data_read.R
strengejacke: I think it's a nice "side-effect" we should allow. I always forget whether functions are names "read" or "load", and had several tries until I could read an rds-file. So I thought `data_read()` is simpler to remember - but this didn't work for a model objects stored in the file. So I thought would be nice if `data_read()` can do this, too. But feel free to turn it into an error if you feel that _data_ read() should only read real data sets.

etiennebacher: No no I'm fine with `data_read` being able to read `rds`. 

What I meant is that you should put it in `.read_unknown()` and not `.post_process_imported_data()` because the latter is called by other `.read_` functions

etiennebacher: not sure my previous message clarifies but I'm fine with the code with your latest changes

strengejacke: Yes, I understood moving the code part. My comment was referring to your suggestion using `.coerce_to_dataframe()`, which errors, while my code just warns

etiennebacher: I thought `tryCatch` would capture the error anyway

strengejacke: Yeah, but then we would have a less specific error message:

> "`data` must be a data frame, or an object that can be coerced to a data frame."

IndrajeetPatil:
@strengejacke Running tests in parallel has nothing to do with this. All tests are failing also on the `main` branch now.

You keep blaming and undoing my changes when they have nothing to do with the problems you are facing.

PR #389, opened 2023-03-22
Attempt to fix broken main

IndrajeetPatil:


strengejacke:
Moving everything inside the `testthat()` function resolved this issue. It worked locally, so it indeed seemed to be an issue of parallel testing before, or not? Else, I can't really figure out why tests were failing...

IndrajeetPatil:
The tests were failing also on the main branch which did not have parallel testing turned on. So unlikely this has anything to do with parallel testing. 

I turned on parallel testing again in this PR.

IndrajeetPatil:
But how can parallel testing be an issue when parallel testing was turned off?!

You removed it before merging your PR to the main. And tests started failing on main.

Also, note that parallel testing runs tests in different files in parallel, and not the individual tests. So a single worker will still be executing all tests in test-data-read.R.

That said, I need to look into why the code coverage is workflow is failing with the same tests failing.

strengejacke:
I don't know, but why did parallel tests first failed (as well), and stopped failing after fixing the two testfiles that both used `temp_file`  as variable? Only explanation I have is that these two files were run in parallel.

strengejacke:
This shouldn't change anything, as tests are now more isolated and should work for both, if I'm not mistaken. It should, however, work, when parallel testing is removed and parts of the code moved outside `testthat()` again...

Another explanation might be that creating temp files and writing binary files in tests may cause issues (or may cause issues when not inside `testthat()`)? Difficult to track down...

etiennebacher:
Here's something very strange that causes some (all?) failures.

In `test-convert_na_to.R`, we load `parameters` with `skip_if_not_or_load_if_installed("parameters")`. For some reason, it causes a failure in `test-data_read.R`, at the test "RDS file, no data frame" but only when we run `devtools::test()`, not in other scenario. It seems that loading `parameters` changes the type of the data saved in the tempfile.

Hard to explain so I made a reprex here: https://github.com/etiennebacher/foo

To reproduce:

* clone the repo
* run `devtools::test()`. There's a failure
* comment the line `skip_if_not_or_load_if_installed("parameters")` and it works fine

We run `skip_if_not_or_load_if_installed()` in a `test_that()` call so there should be no communication. I don't understand why this is happening

etiennebacher:
Note that this also happens with `skip_if_not_installed()`, not only `skip_if_not_or_load_if_installed()`

strengejacke:
See https://github.com/easystats/parameters/blob/main/R/bootstrap_model.R

I'm not sure if this was related to _bayestestR_ functions, where we obtain posteriors from models using `as.data.frame()`, and these methods do something similar for frequentist models?

etiennebacher:
That's the issue:
``` r
library(testthat)

test_that("foo 2", {
  print("as.data.frame.lm" %in% methods(as.data.frame))
})
#> [1] FALSE <------------------------------- FINE
#> ── Skip (???): foo 2 ───────────────────────────────────────────────────────────
#> Reason: empty test

test_that("foo 1", {
  skip_if_not_installed("parameters")
})
#> ── Skip (???): foo 1 ───────────────────────────────────────────────────────────
#> Reason: empty test

test_that("foo 2", {
  print("as.data.frame.lm" %in% methods(as.data.frame))
})
#> [1] TRUE <------------------------------- WHY???
#> ── Skip (???): foo 2 ───────────────────────────────────────────────────────────
#> Reason: empty test
```

IndrajeetPatil:
Why on earth do we have a data frame method for lm objects?!

Isn't that exactly what _parameters_ and _broom_ are for?

I am glad Daniel had a hunch here. It would have taken ages to track this down otherwise.

etiennebacher:
I think there are two problems here:

1. this data.frame method in `parameters`: should probably be removed
2. the question about methods leaking between `test_that()` calls. I made an [SO post](https://stackoverflow.com/questions/75822979/why-do-packages-loaded-inside-test-that-provide-their-methods-outside-of-te), let's see how it goes and if there are no answers I'll make an issue on `testthat` repo

PR #392, opened 2023-03-23
Clean up `data_read()` tests

IndrajeetPatil:


IndrajeetPatil:
Interesting:

Tests
- pass with `devtools::test()`, but
-  fail with `devtools::check()`

IndrajeetPatil:
As @etiennebacher had already pointed out, I can confirm in https://github.com/easystats/parameters/pull/862 that removing `as.data.frame` methods from `{parameters}` fixes these tests.

etiennebacher:
The R CMD check with macOS fails with the same error, right? There are much less failures than before though

IndrajeetPatil:
I am going to merge this. I know it doesn't fix the issue, but the test clean-up makes this PR already worth a merge.

IndrajeetPatil:
> And I'd still like to know how to avoid this namespace contamination between `test_that()` calls. I saw `withr::with_package()` but doesn't seem to work

Yeah, I would also like to figure it out. I will look into it today.

One thing is for sure: our practice of just attaching every package we need for tests to the search path is problematic. We should attach only what is needed to avoid S3 method overwrites, etc.

etiennebacher:
> One thing is for sure: our practice of just attaching every package we need for tests to the search path is problematic. We should attach only what is needed to avoid S3 method overwrites, etc.

Not sure this would solve it. Even calling the function directly with `parameters::` loads the method in the namespace:
``` r
library(testthat)

test_that("foo 1", {
  print("as.data.frame.lm" %in% methods(as.data.frame))
})
#> [1] FALSE
#> ── Skip (???): foo 1 ───────────────────────────────────────────────────────────
#> Reason: empty test

test_that("foo 2", {
  print(parameters::model_parameters)
})
#> function (model, ...) 
#> {
#>     UseMethod("model_parameters")
#> }
#> <bytecode: 0x00000288d64f6570>
#> <environment: namespace:parameters>
#> ── Skip (???): foo 2 ───────────────────────────────────────────────────────────
#> Reason: empty test

test_that("foo 3", {
  print("as.data.frame.lm" %in% methods(as.data.frame))
})
#> [1] TRUE
#> ── Skip (???): foo 3 ───────────────────────────────────────────────────────────
#> Reason: empty test
```

PR #397, opened 2023-03-27
Complete snapshot tests

etiennebacher:


IndrajeetPatil (approved):
Thanks! Yes, I agree, snapshot testing makes sense here.

Daniel is going to be like...

grumpy-ew

PR #406, opened 2023-04-18
`weighted` functions should better deal with `NA` values.

### LINKED ISSUE #394: `weighted` functions should better deal with `NA` values.
Currently, there's no possibility to clean the variable and weights from `NA`s in _datawizard_ functions. I think we should to pairwise deletion of cases with missing value, so we have only complete cases left, and then compute the weighted statistic - as default behaviour.

``` r
x <- c(1, 2, 3, NA, 4)
w <- c(1, NA, 0.5, 0.8, 1.2)

# current behaviour
datawizard::weighted_mean(x, w)
#> [1] NA
# desired behaviour
report:::.weighted_mean(x, w)
#> [1] 2.703704

# current behaviour
datawizard::weighted_median(x, w)
#> [1] 3
# desired behaviour
report:::.weighted_median(x, w)
#> [1] 3

# current behaviour
datawizard::weighted_sd(x, w)
#> [1] NA
# desired behaviour
report:::.weighted_sd(x, w)
#> [1] 1.708622
```

<sup>Created on 2023-03-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

strengejacke:
Fixes #394

### DIFF
--- a/R/weighted_mean_median_sd_mad.R
+++ b/R/weighted_mean_median_sd_mad.R
@@ -25,6 +25,13 @@ weighted_mean <- function(x, weights = NULL, verbose = TRUE, ...) {
     return(mean(x, na.rm = TRUE))
   }
 
+  # remove missings
+  x[is.na(weights)] <- NA
+  weights[is.na(x)] <- NA
+
+  weights <- stats::na.omit(weights)
+  x <- stats::na.omit(x)
+
   stats::weighted.mean(x, weights, na.rm = TRUE)
 }
 
@@ -69,6 +76,13 @@ weighted_sd <- function(x, weights = NULL, verbose = TRUE, ...) {
   if (!.are_weights(weights) || !.validate_weights(weights, verbose)) {
     return(stats::sd(x, na.rm = TRUE))
   }
+
+  # remove missings
+  x[is.na(weights)] <- NA
+  weights[is.na(x)] <- NA
+
+  weights <- stats::na.omit(weights)
+  x <- stats::na.omit(x)
 
   weights1 <- weights / sum(weights)
   center <- sum(weights1 * x)


R/weighted_mean_median_sd_mad.R
```diff
@@ -25,6 +25,13 @@ weighted_mean <- function(x, weights = NULL, verbose = TRUE, ...) {
     return(mean(x, na.rm = TRUE))
   }
 
+  # remove missings
+  x[is.na(weights)] <- NA
+  weights[is.na(x)] <- NA
+
+  weights <- stats::na.omit(weights)
+  x <- stats::na.omit(x)
```
IndrajeetPatil, line 33: We are using this same block of code twice. Maybe turn it into a small utility function?

R/weighted_mean_median_sd_mad.R
```diff
@@ -102,3 +104,27 @@ weighted_mad <- function(x, weights = NULL, constant = 1.4826, verbose = TRUE, .
...
+    # remove Inf
+    x[is.infinite(x)] <- NA
+    weights[is.infinite(weights)] <- NA
+    flag <- TRUE
+  }
+
+  if (anyNA(x) || anyNA(weights)) {
+    # remove missings
+    x[is.na(weights)] <- NA
+    weights[is.na(x)] <- NA
```
etiennebacher, line 120: These two lines are not necessary since they're already `NA`

R/weighted_mean_median_sd_mad.R
```diff
@@ -3,10 +3,11 @@
 #' @inheritParams stats::weighted.mean
 #' @inheritParams stats::mad
 #' @param weights A numerical vector of weights the same length as `x` giving
-#'   the weights to use for elements of `x`.
+#'   the weights to use for elements of `x`. If `weights = NULL`, `x` is passed
+#' to the non-weighted function.
 #' @param verbose Show warning when `weights` are negative?
 #'
-#' If `weights = NULL`, `x` is passed to the non-weighted function.
+#' @note Missing and infinite values are silently removed from `x` and `weights`.
```
etiennebacher, line 10: Shouldn't there be a `na.rm` argument?

R/weighted_mean_median_sd_mad.R
strengejacke: You may have different observations with `NA` in `x` and `weights`. This ensures something like `complete.cases()` for `x` and `weights` when we use `na.omit()`.

R/weighted_mean_median_sd_mad.R
strengejacke: Not sure, we did not had it before.

R/weighted_mean_median_sd_mad.R
etiennebacher: ah right, I didn't notice `x` and `weights` were inverted

NEWS.md
```diff
@@ -6,6 +6,12 @@ BREAKING CHANGES
   `object_has_names()`, `object_has_rownames()`, `is_empty_object()`,
   `compact_list()`, `compact_character()`.
 
+CHANGES
+
+* Weighted-functions (`weighted_sd()`, `weighted_mean()`, ...) gain an `na.rm`
+  argument, to remove or keep missing and infinite values. By default,
+  `na.rm = TRUE`, i.e. missing and infinite values are removed by default.
```
etiennebacher, line 13: I think the default should be `FALSE`

etiennebacher: Ok I'm fine with it

Issue #408, opened 2023-04-23
dw_transformer hits again


DominiqueMakowski:
``` r
library(tidyverse)

iris |> 
  group_by(Species) |> 
  mutate(Sepal.Length = datawizard::normalize(Sepal.Length))
#> Error in `mutate()`:
#> ! Problem while computing `Sepal.Length =
#>   datawizard::normalize(Sepal.Length)`.
#> Caused by error:
#> ! `Sepal.Length` must return compatible vectors across groups.
#> ℹ Result type for group 1 (Species = setosa): <dw_transformer>.
#> ℹ Result type for group 2 (Species = versicolor): <dw_transformer>.
```

<sup>Created on 2023-04-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Current work around is to add `as.numeric()` around the normalized score, but it's a bit annoying nonetheless and I'm not sure how to address that neatly :(

strengejacke:
I suggest using `transform()` or opening an issue at the dplyr-repo. 😬 

```r
library(tidyverse)

iris |> 
  group_by(Species) |> 
  transform(Sepal.Length = datawizard::normalize(Sepal.Length))
```

mattansb:
`transform()` doesn't respect the `group_by()` action.

I think this error is because how the tidyverse treats vectors ([it's complicated](https://vctrs.r-lib.org/)).

Solution 1:

``` r
library(datawizard)
#> Warning: package 'datawizard' was built under R version 4.2.3

iris |> 
  dplyr::group_by(Species) |> 
  normalize(select = Sepal.Length)
#> # A tibble: 150 × 5
#> # Groups:   Species [3]
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#>  *        <dbl>       <dbl>        <dbl>       <dbl> <fct>  
#>  1       0.533          3.5          1.4         0.2 setosa 
#>  2       0.400          3            1.4         0.2 setosa 
#>  3       0.267          3.2          1.3         0.2 setosa 
#>  4       0.200          3.1          1.5         0.2 setosa 
#>  5       0.467          3.6          1.4         0.2 setosa 
#>  6       0.733          3.9          1.7         0.4 setosa 
#>  7       0.200          3.4          1.4         0.3 setosa 
#>  8       0.467          3.4          1.5         0.2 setosa 
#>  9       0.0667         2.9          1.4         0.2 setosa 
[...]

PR #409, opened 2023-04-24
dw_transformer hits again

strengejacke:
#408

This PR doesn't fix 408, but we don't need the class attribute for data frame methods, so we can remove it and thereby reduce possible issues we do not yet anticpate.

### DIFF
--- a/R/data_rescale.R
+++ b/R/data_rescale.R
@@ -130,7 +130,10 @@ rescale.numeric <- function(x,
   attr(out, "new_max") <- new_max
   attr(out, "range_difference") <- max_value - min_value
   attr(out, "to_range") <- c(new_min, new_max)
-  class(out) <- c("dw_transformer", class(out))
+  # don't add attribute when we call data frame methods
+  if (!isFALSE(dot_args$add_transform_class)) {
+    class(out) <- c("dw_transformer", class(out))
+  }
 
   out
 }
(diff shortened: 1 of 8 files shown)


R/data_rescale.R
```diff
@@ -130,7 +130,10 @@ rescale.numeric <- function(x,
   attr(out, "new_max") <- new_max
   attr(out, "range_difference") <- max_value - min_value
   attr(out, "to_range") <- c(new_min, new_max)
-  class(out) <- c("dw_transformer", class(out))
+  # don't add attribute when we call data frame methods
+  if (!isFALSE(dot_args$add_transform_class)) {
```
etiennebacher, line 134: `!isFALSE` -> `isTRUE` ?

R/normalize.R
```diff
@@ -159,7 +159,10 @@ normalize.numeric <- function(x, include_bounds = TRUE, verbose = TRUE, ...) {
   attr(out, "min_value") <- min_value
   attr(out, "vector_length") <- vector_length
   attr(out, "range_difference") <- range_difference
-  class(out) <- c("dw_transformer", class(out))
+  # don't add attribute when we call data frame methods
+  if (!isFALSE(dot_args$add_transform_class)) {
```
etiennebacher, line 163: same

etiennebacher (approved):
~~Minor comment, otherwise~~ LGTM, thanks!

PR #410, opened 2023-04-28
append for `reverse()`

strengejacke:


strengejacke:
``` r
library(datawizard)
test <- data.frame(
  x = 1:5,
  y = c(3, 8, 2, 5, 1)
)

reverse_scale(test, select = "x")
#>   x y
#> 1 5 3
#> 2 4 8
#> 3 3 2
#> 4 2 5
#> 5 1 1

reverse_scale(test, select = "x", append = TRUE)
#>   x y x_r
#> 1 1 3   5
#> 2 2 8   4
#> 3 3 2   3
#> 4 4 5   2
#> 5 5 1   1

reverse_scale(test, append = TRUE)
#>   x y x_r y_r
#> 1 1 3   5   6
#> 2 2 8   4   1
#> 3 3 2   3   7
#> 4 4 5   2   4
#> 5 5 1   1   8
```

<sup>Created on 2023-04-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

etiennebacher:
I'm looking at the test failures and I don't understand why `.process_std_args()` has both `select` and `exclude` because the variable selection is done beforehand in `.select_nse()` so these shouldn't be needed.

strengejacke:
`process_std_args()` indeed does some more things only needed for `center()` and `standardize()`, so I extracted the relevant code for just appending variables and added the `.process_appned()` helper function, which is now used in all other transformation functions... Let's hope everything still works 😬

strengejacke:
Will look into test failures later

etiennebacher (approved):
Looks good to me, thanks @strengejacke

PR #411, opened 2023-04-30
Several improvements

strengejacke:
- selecte NSE handles group-var from grouped df
- add `append` to more methods (`normalize()`, ...)
- where we have an `append` argument, the `.grouped_df` methods need to set `append = FALSE` when calling regular data frame methods
- removes unnecessary formula-evaluation in `normalize()`
- fixed bug in `center.group_df()` when `force = TRUE`


strengejacke:
Need to add tests for `normalize.data.frame(append = TRUE)` and `rescale.data.frame(append = TRUE)`.

R/data_rescale.R
```diff
@@ -202,6 +222,21 @@ rescale.data.frame <- function(x,
     verbose = verbose
   )
 
+  # when we append variables, we call ".pro6cess_append()", which will
```
etiennebacher, line 225:
```suggestion
  # when we append variables, we call ".process_append()", which will
```

tests/testthat/test-normalize.R
```diff
@@ -167,6 +173,24 @@ test_that("normalize (grouped data)", {
   expect_identical(datawizard, manual)
 })
 
+test_that("normalize (grouped data)z, with append", {
```
etiennebacher, line 176:
```suggestion
test_that("normalize (grouped data), with append", {
```

etiennebacher (approved):
LGTM, thanks!

PR #413, opened 2023-05-02
`remove_na` vs. `na.rm`

### LINKED ISSUE #412: `remove_na` vs. `na.rm`
We currently mix both argument names, `remove_na` vs. `na.rm`. I suggest we decide on one and deprecate the other one.

strengejacke:
Fixes #412

### DIFF
--- a/R/descriptives.R
+++ b/R/descriptives.R
@@ -75,7 +75,7 @@ coef_var.default <- function(x, verbose = TRUE, ...) {
 #'   each end of `x` before the mean and standard deviation (or other measures)
 #'   are computed. Values of `trim` outside the range of (0 to 0.5) are taken
 #'   as the nearest endpoint.
-#' @param na.rm Logical. Should `NA` values be removed before computing (`TRUE`)
+#' @param remove_na Logical. Should `NA` values be removed before computing (`TRUE`)
 #'   or not (`FALSE`, default)?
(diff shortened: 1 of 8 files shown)


R/descriptives.R
```diff
@@ -110,14 +111,20 @@ coef_var.default <- function(x, verbose = TRUE, ...) {
 #' @export
 coef_var.numeric <- function(x, mu = NULL, sigma = NULL,
                              method = c("standard", "unbiased", "median_mad", "qcd"),
-                             trim = 0, na.rm = FALSE, n = NULL, ...) {
+                             trim = 0, remove_na = FALSE, n = NULL, na.rm = FALSE, ...) {
+  # TODO: remove deprecated argument later
+  if (!missing(na.rm)) {
+    # TODO: add deprecation warning in a later update
+    remove_na <- na.rm
+  }
```
etiennebacher, line 119: why not add the deprecation message now? that way we can use `remove_na` in other easystats packages as soon as we see the message

etiennebacher: bump

strengejacke: I prefer a "smoother" transition to deprecation, but I don't mind adding the warning now.

etiennebacher: We need a way to tell the users to use `remove_na` instead of `na.rm` so I guess the sooner the better

etiennebacher (approved):
LGTM, thanks @strengejacke

PR #414, opened 2023-05-02
data_select(): add option to evaluate a function passed to `select`

### LINKED ISSUE #407: data_select(): add option to evaluate a function passed to `select`
Trying to fix a bug in modelbased: but it basically fails because the function is not evaluated:

``` r
model <- lm(Petal.Length ~ Petal.Width, data=iris)
datawizard::data_select(iris, insight::find_predictors(model, effects = "fixed", flatten = TRUE))
#> Error in switch(type, `:` = .select_seq(x, data, ignore_case, regex, verbose), : EXPR must be a length 1 vector
```

<sup>Created on 2023-04-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I understand the need to catch unevaluated function calls (for stuff like `as.numeric` to work), but we should either come up with a smarter to know when to evaluate vs. not, or add an argument `force_eval` or something to force it to evaluate

strengejacke:
Fixes #407

### DIFF
--- a/R/select_nse.R
+++ b/R/select_nse.R
@@ -201,6 +201,13 @@
       }
     )
 
+    # when "x" is a function arg which is itself a function call to evaluate,
+    # .dynGet can return "x" infinitely so we try to evaluate this arg
+    # see #414
+    if (!is.null(new_expr) && insight::safe_deparse(new_expr) == "x") {
+      new_expr <- .dynEval(x, inherits = FALSE, minframe = 0L, remove_n_top_env = 4)
+    }
+
     if (is_select_helper) {
       new_expr <- str2lang(unlist(new_expr, use.names = FALSE))
       out <- .eval_expr(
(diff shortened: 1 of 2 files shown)


R/select_nse.R
```diff
@@ -47,6 +47,11 @@
     verbose = verbose
   )
 
+  # sanity check, if ".eval_expr" returns a column name, and no index (see #407)
+  if (is.character(selected) && all(selected %in% columns)) {
+    selected <- which(selected %in% columns)
+  }
+
```
strengejacke, line 54:
```r
library(datawizard)
model <- lm(Petal.Length ~ Petal.Width, data=iris)
datawizard::data_select(iris, insight::find_predictors(model, effects = "fixed", flatten = TRUE))
```

@etiennebacher in Dom's example above, `.eval_expr()` now returns a column name, no column index. I added the sanity check here, but maybe it's better placed in `.eval_expr()` - However, I wasn't sure if a character vector can be a valid return value for `.eval_expr()`, so I didn't want to place the code there. Maybe you can check, and probably move it to the best place?

etiennebacher:
Thanks, I changed a bit the code to leave `.select_nse()` unchanged.

Before merging, I'd like to fix this strange issue (which is why I skipped CI, I know it will error):
``` r
library(datawizard)

model <- lm(Petal.Length ~ Petal.Width, data = iris)

# works fine with arg named "i"
foo <- function(i) {
  head(data_select(iris, i))
}
foo(insight::find_predictors(model, effects = "fixed", flatten = TRUE))
#>   Petal.Width
#> 1         0.2
#> 2         0.2
#> 3         0.2
#> 4         0.2
#> 5         0.2
#> 6         0.4

# error with arg named "x"
foo <- function(x) {
  head(data_select(iris, x))
}
foo(insight::find_predictors(model, effects = "fixed", flatten = TRUE))
#> NULL
```

PR #417, opened 2023-05-15
Implement `rowid_as_column()`

### LINKED ISSUE #416: Add select helpers to `data_to_wide` and friends
I am currently trying to use `datawizard::data_to_wide()` in another PR but it seems like it is throwing an error because of differing number of rows, which seems cannot be fixed by specifying columns to be selected because the argument has not been implemented for those functions so far. Reprex:

``` r
datawizard::data_to_wide(
  iris,
  values_from = "Sepal.Length",
  names_from = "Species",
  select = c("Sepal.Length", "Species")
)
#> Error in (function (..., row.names = NULL, check.rows = FALSE, check.names = TRUE, : arguments imply differing number of rows: 147, 145, 144
```

<sup>Created on 2023-05-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The workaround is to run `datawizard::data_select()` on the data frame before hand, but it would be nice if it was integrated.

etiennebacher:
* Add `rowid_as_column()`
* split rownames and colnames functions into 2 separate R files, same for tests files
* fix an example that requires `gamm4`

Close #416


etiennebacher:
Thank you @strengejacke, I had completely forgotten attributes. I think it would be better to keep them, I added some tests for this.

I also added some extra checks for the name of the new column created by `rownames_as_column()` and `rowid_as_column()`

R/utils-rows.R
```diff
@@ -0,0 +1,125 @@
...
+      paste0("There is already a variable named `", var, "` in your dataset.")
+    )
+  }
+  # works only for dplyr >= 0.8.0
+  grps <- attr(x, "groups", exact = TRUE)
+  grps <- grps[[".rows"]]
+
+  for (i in seq_len(length(grps))) {
+    x[grps[[i]], var] <- seq_len(length(grps[[i]]))
+  }
```
strengejacke, line 118: I think you can replace both `seq_len(length())` with `seq_along()`.

strengejacke: But I wonder if this is really what we want? We would have duplicated row id's here, i.e. each "group" starts with ID 1 to ...

etiennebacher: I thought `tibble::rowid_to_column()` did the same way as I do here for groups but apparently it just ignores them. I implemented this group behavior to fix the example in #416 but maybe I should remove it

etiennebacher: I'd still like to have an easy way to create row index per group that would be equivalent to:
``` r

iris |> 
  sample_n(10) |> 
  group_by(Species) |> 
  mutate(id = row_number()) |> 
  ungroup()
#> # A tibble: 10 × 6
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species       id
#>           <dbl>       <dbl>        <dbl>       <dbl> <fct>      <int>
#>  1          5.8         2.8          5.1         2.4 virginica      1
#>  2          4.7         3.2          1.6         0.2 setosa         1
#>  3          6.4         2.9          4.3         1.3 versicolor     1
#>  4          6           2.7          5.1         1.6 versicolor     2
#>  5          7           3.2          4.7         1.4 versicolor     3
#>  6          6.3         2.8          5.1         1.5 virginica      2
#>  7          4.4         3            1.3         0.2 setosa         2
#>  8          7.7         2.8          6.7         2   virginica      3
#>  9          5.7         3            4.2         1.2 versicolor     4
#> 10          6.1         2.6          5.6         1.4 virginica      4
```
but I don't know if we want to implement a new function for this or keep it in a grouped method for `rowid_as_column()`

strengejacke: We can keep this for the grouped_df method, I just wanted to be sure this is intentional.

etiennebacher: I added some details about this in the docs and in NEWS

etiennebacher:
I don't understand the macOS errors, `testthat` doesn't show any issues. @IndrajeetPatil I think you have a Mac, can you try to replicate those?

rempsyc:
I noticed that the `poorman` approach supports unequal groups (e.g., using `mtcar`'s `cyl` as group, or removing one row from iris). Can you check if this new approach works with unequal n as well? That would be nice.

rempsyc:
And the macos error is unusual, and couldn't find any match on Google, hum.

``` r
Shifting non-numeric variables is not possible.
  Try using 'to_numeric()' and specify the 'lowest' argument.
```

rempsyc:
And nice for the `grouped_df` method, I think when merged this can close #416.

etiennebacher:
> I noticed that the poorman approach supports unequal groups

See comment in #416. 

> And the macos error is unusual, and couldn't find any match on Google, hum.

These are just messages given by `datawizard`, they are not errors. The source of the error is not shown

etiennebacher:
@strengejacke the errors are not due to this PR, they also appear in other PR (for example #422). So if you don't have other comments can you approve this?

strengejacke (approved):

Issue #423, opened 2023-05-25
datawizard versions of separate() and unite()


jmgirard:
I'm hoping to retool my packages soon to use datawizard instead of dplyr and tidyr. However, a few things I need that don't seem to exist yet are datawizard versions of separate() and unite() from tidyr. Any plans to create these?

etiennebacher:
I think `unite()` should be fairly easy to implement, but `separate()` is a bit trickier because of the arguments `fill` and `extra`. I started working on implementing `separate()` in `poorman` (cf https://github.com/nathaneastwood/poorman/pull/107) but it's still WIP (and I don't know when I'll get back at it).

strengejacke:
``` r
library(datawizard)

d <- data.frame(
  x = c(NA, 1:3),
  y = c(letters[1:3], NA_character_),
  z = 6:9,
  m = c("X", NA_character_, "Y", "Z")
)
d
#>    x    y z    m
#> 1 NA    a 6    X
#> 2  1    b 7 <NA>
#> 3  2    c 8    Y
#> 4  3 <NA> 9    Z

data_unite(d, new_column = "xyz")
#>         xyz
#> 1  NA_a_6_X
#> 2  1_b_7_NA
#> 3   2_c_8_Y
#> 4  3_NA_9_Z
data_unite(d, new_column = "xyz", remove_na = TRUE)
#>        xyz
#> 1    a_6_X
#> 2    1_b_7
#> 3  2_c_8_Y
#> 4    3_9_Z

data_unite(d, new_column = "x")
#>           x
#> 1  NA_a_6_X
#> 2  1_b_7_NA
#> 3   2_c_8_Y
#> 4  3_NA_9_Z
data_unite(d, new_column = "x", remove_na = TRUE)
#>          x
#> 1    a_6_X
#> 2    1_b_7
#> 3  2_c_8_Y
#> 4    3_9_Z

data_unite(d, new_column = "x", append = TRUE)
#> The name for `new_column` already exists as variable name in the data.
#>   This variable will be replaced by `new_column`.
#>           x    y z    m
#> 1  NA_a_6_X    a 6    X
#> 2  1_b_7_NA    b 7 <NA>
#> 3   2_c_8_Y    c 8    Y
#> 4  3_NA_9_Z <NA> 9    Z
data_unite(d, new_column = "x", remove_na = TRUE, append = TRUE)
#> The name for `new_column` already exists as variable name in the data.
#>   This variable will be replaced by `new_column`.
[...]

PR #424, opened 2023-05-25
datawizard version of `unite()`

strengejacke:
Partly addresses #423


R/data_unite.R
```diff
@@ -0,0 +1,108 @@
...
+  if (is.null(select) || length(select) <= 1) {
+    insight::format_error(
+      "At least two columns in `select` are required for `data_unite()`."
+    )
+  }
+
+  # transpose and unite
+  d <- data[select]
+  out <- as.data.frame(do.call(rbind, lapply(data_transpose(d, verbose = FALSE), paste0, collapse = separator)))
+  colnames(out) <- new_column
```
etiennebacher, line 78: You can use this instead (much faster):
```r
out[[new_column]] <- do.call(paste, c(data[, select], sep = separator))
```

R/data_unite.R
```diff
@@ -0,0 +1,108 @@
...
+
+  # remove missings
+  if (remove_na) {
+    out[[new_column]] <- gsub("NA_", "", out[[new_column]], fixed = TRUE)
+    out[[new_column]] <- gsub("_NA", "", out[[new_column]], fixed = TRUE)
+  }
```
etiennebacher, line 84:
```r
out[[new_column]] <- gsub(paste0(separator, "NA", separator), "", out[[new_column]], fixed = TRUE)
out[[new_column]] <- gsub(paste0("^NA", separator), "", out[[new_column]], fixed = TRUE)
out[[new_column]] <- gsub(paste0(separator, "NA$"), "", out[[new_column]], fixed = TRUE)
```

R/data_unite.R
```diff
@@ -0,0 +1,108 @@
...
+    data[select] <- NULL
+  }
+
+  # overwrite?
+  if (new_column %in% colnames(data) && verbose) {
+    insight::format_alert(
+      "The name for `new_column` already exists as variable name in the data.",
+      "This variable will be replaced by `new_column`."
+    )
+  }
```
etiennebacher, line 97: this check should come before uniting the columns

R/data_unite.R
```diff
@@ -0,0 +1,108 @@
...
+
+  # overwrite or append
+  if (new_column %in% colnames(data)) {
+    data[[new_column]] <- out[[new_column]]
+  } else {
+    data <- cbind(data, out)
+  }
```
etiennebacher, line 104: This shouldn't be needed

R/data_unite.R
```diff
@@ -0,0 +1,108 @@
...
+
+  # fin
+  data
```
etiennebacher, line 107: does it preserve attributes of untouched columns?

R/data_unite.R
strengejacke: I'm not sure we should have `gsub(paste0(separator, "NA", separator)`, which will turn `a_NA_1` into `a1` instead of `a_1`. And we don't need regex (i.e. `$` or `^`), as these cases should be already covered.

R/data_unite.R
strengejacke: If old columns should be removed, they will not be overwritten. That's why I decided to put this check here. Else, we get a "false positive" warning.

R/data_unite.R
strengejacke: You mean `cbind()`? why not?

R/data_unite.R
strengejacke: Must check.

strengejacke: Yes, works

R/data_unite.R
etiennebacher: Ah yeah I didn't notice it was just a message, I thought it was an error. So its position doesn't really matter

R/data_unite.R
etiennebacher: Agreed for the first point but for the second one we need `$` and `^` to avoid removing text in caps:

``` r
library(datawizard)

d <- data.frame(x = "COUNTRY", y = "NATION")

data_unite(d, "foo", remove_na = TRUE)
#>         x      y         foo
#> 1 COUNTRY NATION COUNTRYTION
```

etiennebacher: Actually even for the first point I would add:

```r
out[[new_column]] <- gsub(paste0(separator, "NA", separator), "_", out[[new_column]], fixed = TRUE)
```
(we add a `_` back)

R/data_unite.R
etiennebacher: Doesn't `data[[new_column]] <- out[[new_column]]` just work? without the if-else

R/data_unite.R
strengejacke: Good point!

R/data_unite.R
```diff
@@ -0,0 +1,106 @@
...
+  )
+
+  if (is.null(select) || length(select) <= 1) {
+    insight::format_error(
+      "At least two columns in `select` are required for `data_unite()`."
+    )
+  }
+
+  # unite
+  out[[new_column]] <- do.call(paste, c(d[select], sep = separator))
```
etiennebacher, line 76: my suggestion earlier was incomplete, need to define `out` and replace `d` by `data` here

strengejacke: Just realized this when writing tests ;-)

R/data_unite.R
```diff
@@ -0,0 +1,111 @@
+#' @title Unite ("merge") multiple variables
+#' @name data_unite
+#'
+#' @description
+#' Merges values of multiple variables per observation into one new variable.
```
etiennebacher, line 5:
```suggestion
#' Merge values of multiple variables per observation into one new variable.
```

R/data_unite.R
```diff
@@ -0,0 +1,111 @@
...
+#'
+#' @param data A data frame.
+#' @param new_column The name of the new column, as a string.
+#' @param separator Separator, as string, to use between values.
```
etiennebacher, line 9:
```suggestion
#' @param separator A character to use between values.
```

R/data_unite.R
```diff
@@ -0,0 +1,111 @@
...
+#' @param append Logical, if `FALSE` (default), removes original columns that
+#' were united. If `TRUE`, all columns are preserved and the new column is
+#' appended to the data frame.
```
etiennebacher, line 12: Actually I think it would make more sense to go the other way and to have an argument `remove` that is `TRUE` by default (just to stick to `tidyr::unite()`)

etiennebacher (commented):
A few comments, mostly on the argument `append`. 

Otherwise once the description and news are updated I think this is good to go (except if you want to add the `separate()` version as well?)

R/data_unite.R
strengejacke: I thought about this, too, but the logic in our package is to `append` (or not to). We can _additionally_ have a `remove` argument that simply defaults to `!append`?

strengejacke:
>  (except if you want to add the separate() version as well?)

Maybe in a separate PR.

R/data_unite.R
etiennebacher: It's better to only have one of the two so we can keep `append` for consistency with the rest

PR #438, opened 2023-06-26
recode_into can skip overwriting former recodes

strengejacke:
`recode_into()` by defaults overwrite former recodings when multiple recode patterns apply. This PR adds an `overwrite` argument that allows users to skip this behaviour, preserving already recoded cases and not overwrite them.

``` r
library(datawizard)
x <- 1:30

recode_into(
  x > 1 ~ "a",
  x > 10 & x <= 15 ~ "b",
  default = "c"
)
#> Warning: Several recode patterns apply to the same cases. Some of the already
#>   recoded cases will be overwritten with new values again (e.g. pattern 2
#>   overwrites the former recode of case 1).
#>   Please check if this is intentional!
#>  [1] "c" "a" "a" "a" "a" "a" "a" "a" "a" "a" "b" "b" "b" "b" "b" "a" "a" "a" "a"
#> [20] "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a"

recode_into(
  x > 1 ~ "a",
  x > 10 & x <= 15 ~ "b",
  default = "c",
  overwrite = FALSE
)
#> Warning: Several recode patterns apply to the same cases. Some of the already
#>   recoded cases will not be altered by later recode patterns. (e.g.
#>   pattern 2 also matches the former recode of case 1).
#>   Please check if this is intentional!
#>  [1] "c" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a"
#> [20] "a" "a" "a" "a" "a" "a" "a" "a" "a" "a" "a"
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 0.8.0.1
+Version: 0.8.0.2
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0003-1995-6531", Twitter = "@patilindrajeets")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # datawizard (devel)
 
+CHANGES
+
+* `recode_into()` gains an `overwrite` argument to skip overwriting already
+  recoded cases when multiple recode patterns apply to the same case.
+
 BUG FIXES
 
 * Fixed issues in `data_write()` when writing labelled data into SPSS format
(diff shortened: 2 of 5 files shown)


strengejacke:
``` r
library(datawizard)

x <- 1:10

# default behaviour, second patterns overwrites
# formerly recoded values > 5 into 2
recode_into(
  x >= 3 & x <= 7 ~ 1,
  x > 5 ~ 2,
  default = 0,
  overwrite = TRUE,
  verbose = FALSE
)
#>  [1] 0 0 1 1 1 2 2 2 2 2

# behaviour when "overwrite = FALSE", second patterns does not overwrite
# formerly recoded values from 3 to 7, and only changes values > 7
recode_into(
  x >= 3 & x <= 7 ~ 1,
  x > 5 ~ 2,
  default = 0,
  overwrite = FALSE,
  verbose = FALSE
)
#>  [1] 0 0 1 1 1 1 1 2 2 2
```

<sup>Created on 2023-06-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #442, opened 2023-07-01
`standardize()` fails in some cases

### LINKED ISSUE #441: `standardize()` fails in some cases
Cf https://github.com/easystats/report/issues/379

``` r
library(datawizard)

data(sleep)
d <- data_modify(sleep, group = as.integer(group) - 1L)
d_wide <- data_to_wide(
  d,
  names_from = "group",
  values_from = "extra",
  names_prefix = "group"
)

model1 <- lm(d_wide$group0 - d_wide$group1 ~ 1)
standardize(model1)
#> Warning: Using `$` in model formulas can produce unexpected results. Specify your
#>   model using the `data` argument instead.
#>   Try: group0 - group1 ~ 1,
#>   data =

#> Warning: Using `$` in model formulas can produce unexpected results. Specify your
#>   model using the `data` argument instead.
#>   Try: group0 - group1 ~ 1,
#>   data =

#> Warning: Using `$` in model formulas can produce unexpected results. Specify your
#>   model using the `data` argument instead.
#>   Try: group0 - group1 ~ 1,
#>   data =
#> 
#> Call:
#> lm(formula = d_wide$group0 - d_wide$group1 ~ 1, data = data_std)
#> 
#> Coefficients:
#> (Intercept)  
#>       -1.58

model1 <- lm(group0 - group1 ~ 1, data = d_wide)
standardize(model1)
#> Error in eval(predvars, data, env): objet 'group0' introuvable
#> Error: Unable to refit the model with standardized data.
#>   Try instead to standardize the data (standardize(data)) and refit the
#>   model manually.
```

<sup>Created on 2023-07-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

strengejacke:
Fixes #441

### DIFF
--- a/R/standardize.models.R
+++ b/R/standardize.models.R
@@ -84,7 +84,7 @@ standardize.default <- function(x,
     weights = weights,
     verbose = verbose,
     include_response = include_response,
-    update_expr = stats::update(x, data = data_std),
+    update_expr = str2lang("stats::update(x, data = data_std)"),
     ...
   )
 }
@@ -264,10 +264,10 @@ standardize.default <- function(x,
   on.exit(.update_failed())
 
   if (isTRUE(verbose)) {
-    model_std <- eval(substitute(update_expr))
+    model_std <- eval(update_expr)
   } else {
     utils::capture.output({
-      model_std <- eval(substitute(update_expr))
+      model_std <- eval(update_expr)
     })
   }
 


etiennebacher:
@strengejacke do you want to finish this one before 0.9.0? (I won't release before `insight` anyway)

strengejacke:
I can hardly remember the issue... I would add tests based on the example in the references issue, no?

etiennebacher:
Yes I think so, but based on @mattansb's comment in the related issue I'm not sure we should allow this

mattansb:
What issue are we trying to solve here?

I would argue that the examples in #441 should fail since it is not doing what one probably wants (to standardize the composite score).

strengejacke:
Ok, this PR does not fully solve all issue mentioned in #441, but if I understand, this should also be intended? At least, it provides a small improvement, since it no longer fails for intercept-only models. If test coverage is sufficient, the small changes should not break any code, so we can merge this (or close this, if it seems to risky to address the edge cases, w/o knowing whether this change breaks something elsewhere).

mattansb:
I think it's too risky... I also seems maybe a little silly to do this for intercept only models - since the standardized intercept is always 0!

strengejacke:
@etiennebacher Seems like we - for now - don't plan to do anything regarding this issue, so I guess _datawizard_ is ready for submission. We can keep this issue open to think about if we find a good solution how to deal with this (which can be: we already have the best solution, we add an informative message, ....).

PR #445, opened 2023-08-09
Implement `rowmean_n()`

strengejacke:


DominiqueMakowski:
should we name it "rowmean_n" ?

strengejacke:
> should we name it "rowmean_n" ?

`mean_n()` was in reference to SPSS' `MEAN.n`, but we can name it `rowmean_n()`, or add an alias?

DominiqueMakowski:
too much aliases can be confusing, I'd say go with `rowmean` as it's more R-like (and I don't think we need to be consistent with SPSS naming schemes ^^)

R/rowmean_n.R
```diff
@@ -0,0 +1,107 @@
...
+  if (!all(numeric_columns)) {
+    if (verbose) {
+      insight::format_alert("Only numeric columns are considered for calculation.")
+    }
+    data <- data[numeric_columns]
+  }
+
+  # check if we have a data framme with at least two columns
+  if (ncol(data) < 2) {
+    insight::format_error("`data` must be a data frame with at least two columns.")
```
etiennebacher, line 85:
```suggestion
    insight::format_error("`data` must be a data frame with at least two numeric columns.")
```

R/rowmean_n.R
```diff
@@ -0,0 +1,107 @@
...
+    n <- round(ncol(data) * decimals)
+  }
+
+  # n may not be larger as df's amount of columns
+  if (ncol(data) < n) {
+    insight::format_error("`n` must be smaller or equal to number of columns in data frame.")
+  }
+
+  # row means
+  out <- apply(data, 1, function(x) ifelse(sum(!is.na(x)) >= n, mean(x, na.rm = TRUE), NA))
```
etiennebacher, line 100: `rowMeans()` is much faster than `apply()` + `mean()`: 
``` r
large_df <- data.table::rbindlist(rep(list(mtcars), 10000))
bench::mark(
  apply_mean = apply(large_df, 1, mean),
  rowMeans = rowMeans(large_df),
  iterations = 10
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#> 1 apply_mean    2.02s    2.13s     0.472    61.2MB     15.1
#> 2 rowMeans    17.82ms  27.91ms    34.9      29.3MB     13.9
```

etiennebacher (changes requested):

R/rowmean_n.R
strengejacke: Yes, but we only want to apply rowmeans to rows with at least `n` valid values, so we somehow need to identify those rows.

strengejacke:
``` r
d_mn <- data.frame(
  c1 = c(1, 2, NA, 4),
  c2 = c(NA, 2, NA, 5),
  c3 = c(NA, 4, NA, NA),
  c4 = c(2, 3, 7, 8)
)

test1 <- function(data, n = 2) {
  apply(data, 1, function(x) ifelse(sum(!is.na(x)) >= n, mean(x, na.rm = TRUE), NA))
}

test2 <- function(data, n = 2) {
  rowMeans(data, na.rm = TRUE)
}

bench::mark(
  test1(d_mn),
  test2(d_mn),
  iterations = 1000
)
#> Error: Each result must equal the first result:
#> `test1(d_mn)` does not equal `test2(d_mn)`
```

<sup>Created on 2023-08-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

etiennebacher: You can detect which rows should be NA at the end based on their number of missing values and then use `rowMeans()`:
```r
  # row means
  to_na <- rowSums(is.na(data)) > ncol(data) - n
  out <- rowMeans(data, na.rm = TRUE)
  out[to_na] <- NA
```
With this change:
```r
mtcars2 <- mtcars
for (j in sample(1:ncol(mtcars2), 5)) {
  for (i in sample(1:nrow(mtcars2), 5)) {
    mtcars2[i, j] <- NA
  }
}
large_df <- data.table::rbindlist(rep(list(mtcars2), 10000))
bench::mark(
  rowmean_n(large_df, 7),
  new_rowmean_n(large_df, 7),
  iterations = 10
)

# A tibble: 2 × 13
  expression               min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory    
  <bch:expr>           <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>    
1 rowmean_n(large_df,…   3.22s   3.51s     0.287    61.1MB     8.77    10   306      34.9s <dbl>  <Rprofmem>
2 new_rowmean_n(large… 37.87ms 60.16ms    15.5      61.1MB    10.9     10     7    643.7ms <dbl>  <Rprofmem>
# ℹ 2 more variables: time <list>, gc <list>
```
```

File: R/rowmean_n.R [hunk 3]
```diff
@@ -0,0 +1,107 @@
...
+rowmean_n <- function(data, n, digits = NULL, verbose = TRUE) {
+  # check if data is data frame or matrix
+  if (!is.data.frame(data) && !is.matrix(data)) {
+    insight::format_error("`data` must be a data frame or matrix.")
+  }
+
+  # coerce matrix to data frame
+  if (is.matrix(data)) {
+    data <- as.data.frame(data)
+  }
```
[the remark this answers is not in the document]
etiennebacher: I meant that the first two `if` conditions could be replaced by `.coerce_to_dataframe()`, not just the second one

etiennebacher (approved):

strengejacke:
cool, thanks!

etiennebacher:
thank you ;)

PR #448, opened 2023-08-13
Rename `rowmean_n()` to `row_means()`

### LINKED ISSUE #447: Rename `rowmean_n()`?
From @IndrajeetPatil's comment: https://github.com/easystats/datawizard/commit/6e3182c7eb9602badffb5db94cf3cea8149c193d#commitcomment-124303628

> Sorry, I didn't get to weigh in before, but I personally find the function name very dissatisfying, and even a bit misleading. I think the name should make a reference to the fact that this is relevant for missing values. I know that n is mandatory, and so users may figure it out once they use it mistakenly.
>
> How about renaming it to `rowmean_n_missing()`?

@strengejacke

strengejacke:
Fixes #447


strengejacke:
Still needs some fixing, though.

NEWS.md
```diff
@@ -2,8 +2,8 @@
 
 NEW FUNCTIONS
 
-* `rowmean_n()`, to compute row means if row contains at least `n` non-missing
-  values.
+* `row_means()`, to compute row means, optionally only for those row with at
```
DominiqueMakowski, line 5: why is it plural mean*s* ? doesn't it return just *the* mean?

strengejacke: Like `rowMeans()` from base R, it computes mean_s_ of several rows.

etiennebacher (approved):
@IndrajeetPatil are you ok with this new name?

PR #458, opened 2023-09-12
Implement `data_seek()`

strengejacke:
@easystats/maintainers What do you think about this function? I find it useful when I see a questionnaire of a data set and am looking for particular variables where the pattern I know are present in value or variable labels.

One problem is the name - this functions tries to find variables, but we already have `data_find()` (and `find_columns() as alias), which work differently and also return the extracted variables - `seek_variables()`, on contrast, returns a data frame with a summary of found matches. `find_variables()` is already taken in _insight_.

`data_seek()` is currently an alias.


strengejacke:
btw, this is a slightly modified version of [`sjmisc::find_in_data()`](https://strengejacke.github.io/sjmisc/reference/find_var.html)

strengejacke:
Examples:

``` r
library(datawizard)
# seek variables with "Length" in variable name or labels
seek_variables(iris, "Length")
#>              index       column       labels
#> Sepal.Length     1 Sepal.Length Sepal.Length
#> Petal.Length     3 Petal.Length Petal.Length

# seek variables with "dependency" in names or labels
# column "e42dep" has a label-attribute "elder's dependency"
data(efc)
seek_variables(efc, "dependency")
#>        index column             labels
#> e42dep     3 e42dep elder's dependency

# "female" only appears as value label attribute - default search is in
# variable names and labels only, so no match
seek_variables(efc, "female")
#> [1] index  column labels

# when we seek in all sources, we find the variable "e16sex"
seek_variables(efc, "female", source = "all")
#>        index column         labels
#> e16sex     2 e16sex elder's gender

# typo, no match
seek_variables(iris, "Lenght")
#> [1] index  column labels

# typo, fuzzy match
seek_variables(iris, "Lenght", fuzzy = TRUE)
#>              index       column       labels
#> Sepal.Length     1 Sepal.Length Sepal.Length
#> Petal.Length     3 Petal.Length Petal.Length
```

<sup>Created on 2023-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
As someone that doesn't use labels, I don't expect to use this function over data_find(), but I can see the benefits

etiennebacher:
I like this, it would have been useful for me before. How about `data_search()` for the name?

NAMESPACE
```diff
@@ -283,6 +285,7 @@ export(row_means)
 export(row_to_colnames)
 export(rowid_as_column)
 export(rownames_as_column)
+export(seek_variables)
```
etiennebacher, line 288: I think we should avoid too many aliases, it doesn't seem necessary to have one here (no pref on the one to keep though)

R/seek_variables.R
```diff
@@ -0,0 +1,169 @@
...
+#' @param seek Character vector, indicating where `pattern` is sought. Use one
+#' of more of the following options:
+#'
+#' - `"name"`: searches in column names.
+#' - `"labels"`: searches in variable labels. Only applies when a `label` attribute
+#'   is set for a variable.
+#' - `"values"`: searches in value labels or factor levels. Only applies when a
+#'   `labels` attribute is set for a variable, or if a variable is a factor.
+#' - `"all"`: searches in all of the above.
+#' @param fuzzy Logical, if `TRUE`, "fuzzy matching" (partial and close distance
```
etiennebacher, line 29:
```suggestion
#' @param fuzzy Logical. If `TRUE`, "fuzzy matching" (partial and close distance
```

R/seek_variables.R
```diff
@@ -0,0 +1,169 @@
...
+#' seek_variables(iris, "Lenght", fuzzy = TRUE)
+#' @export
+seek_variables <- function(data, pattern, seek = c("names", "labels"), fuzzy = FALSE) {
+  # check valid args
+  if (!is.data.frame(data)) {
+    insight::format_error("`data` must be a data frame.")
+  }
+
+  # check valid args
+  seek <- intersect(seek, c("names", "labels", "values", "levels", "column_names", "columns", "all"))
```
etiennebacher, line 62: "levels", "column_names", and "columns" are not listed in the possible choices in the docs

NAMESPACE
strengejacke: Agree. The idea for `data_seek()` came later, and that one is more in line with our other function names. I still would prefer `data_seek()` over `data_search()`, though, because _seek_ sounds more like investigation than _search_ to me, but... I'm not a native speaker ;-)

R/seek_variables.R
strengejacke: Yes, that's intentional. I always forget the options, so I added some undocumented aliases for me :-D

R/seek_variables.R
```diff
@@ -0,0 +1,169 @@
...
+        }
+      }
+    }
+    c(pos1, pos2, pos3)
+  }))
+
+  # clean up
+  pos <- unique(pos)
+  # remove -1
+  pos <- pos[which(pos != -1)]
```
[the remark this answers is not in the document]
strengejacke: I think this can no longer happen, it was a leftover from the sjmisc-code.

NAMESPACE
etiennebacher: Well me neither :smile: but of course there's a stackexchange for everything: https://english.stackexchange.com/questions/35507/seek-vs-search

> One fundamental difference is that the object of "seek" is the item you are trying to locate, whereas the object of "search" is the place you are looking

So based on this I would go with `seek` as well since we look for a specific regex

R/seek_variables.R
etiennebacher: alright, might be worth adding a comment about this

strengejacke:
> As someone that doesn't use labels

You could at least use it for factor levels :-)

``` r
datawizard::data_seek(iris, "setosa", seek = "all")
#> index |  column |  labels
#> -------------------------
#>     5 | Species | Species
```

<sup>Created on 2023-09-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

etiennebacher (approved):
Thank you @strengejacke !

PR #467, opened 2023-11-14
rescale() can expand ranges

strengejacke:
Similar to `scales::expand_range()`, `rescale()` can now also expand ranges:

```r
library(datawizard)
x <- seq(5, 15, 2)
x
#> [1]  5  7  9 11 13 15
rescale(x, multiply = 1.1)
#> [1]  4.5  6.7  8.9 11.1 13.3 15.5
#> (original range = 5 to 15)

rescale(x, add = 1)
#> [1]  4.0  6.4  8.8 11.2 13.6 16.0
#> (original range = 5 to 15)

rescale(x, add = -2)
#> [1]  7.0  8.2  9.4 10.6 11.8 13.0
#> (original range = 5 to 15)

rescale(x, add = c(3, 10))
#> [1]  2.0  6.6 11.2 15.8 20.4 25.0
#> (original range = 5 to 15)

rescale(x, add = c(-3, 5))
#> [1]  8.0 10.4 12.8 15.2 17.6 20.0
#> (original range = 5 to 15)


d <- data.frame(x = seq(5, 15, 2), y = seq(5, 15, 2))

# multiply range of each column, expanding by 10%
rescale(d, multiply = 1.1)
#>      x    y
#> 1  4.5  4.5
#> 2  6.7  6.7
#> 3  8.9  8.9
#> 4 11.1 11.1
#> 5 13.3 13.3
#> 6 15.5 15.5

# add 0.5 to each end of range (same as 10% expanding)
rescale(d, add = 0.5)
#>      x    y
#> 1  4.5  4.5
#> 2  6.7  6.7
#> 3  8.9  8.9
#> 4 11.1 11.1
#> 5 13.3 13.3
#> 6 15.5 15.5

# multiply range of x by 1.1 and range of y by 0.5
rescale(d, multiply = list(x = 1.1, y = 0.5))
#>      x    y
#> 1  4.5  7.5
#> 2  6.7  8.5
#> 3  8.9  9.5
#> 4 11.1 10.5
#> 5 13.3 11.5
#> 6 15.5 12.5

# add 0.5 to each end of range of x and subtract 2.5 from each end of range of y
rescale(d, add = list(x = 0.5, y = -2.5))
#>      x    y
#> 1  4.5  7.5
#> 2  6.7  8.5
#> 3  8.9  9.5
#> 4 11.1 10.5
#> 5 13.3 11.5
#> 6 15.5 12.5

# add 1 to lower end of range of x and 3 to upper end of range of x
# add 2 to lower end of range of y and 4 to upper end of range of y
rescale(d, add = list(x = c(1, 3), y = c(2, 4)))
#>      x    y
#> 1  4.0  3.0
#> 2  6.8  6.2
#> 3  9.6  9.4
#> 4 12.4 12.6
#> 5 15.2 15.8
#> 6 18.0 19.0
```

Not sure, though, if `add` should add (subtract) the value to each end of the range? I.e. new range = `c(min - add, max + add)`? Then the range would expand by `2*add`.


strengejacke:
@bwiernik @etiennebacher we could add an alias `expand_range()`, which only has arguments `multiply` and `add` (and not `to`, like `rescale()`), which then calls `rescale()`.

R/data_rescale.R
```diff
@@ -1,15 +1,26 @@
...
 #'
-#' @param to Numeric vector of length 2 giving the new range that the variable will have after rescaling.
-#'   To reverse-score a variable, the range should be given with the maximum value first.
-#'   See examples.
+#' @param to Numeric vector of length 2 giving the new range that the variable
+#'   will have after rescaling. To reverse-score a variable, the range should
+#'   be given with the maximum value first. See examples.
+#' @param multiply If not `NULL`, `to` is ignored and `multiply` will be used,
+#'   giving the factor by which the actual range of `x` should be expanded.
+#'   For example, if a vector range from 5 to 15 and `multiply = 1.1`, the current
```
etiennebacher, line 15:
```suggestion
#'   For example, if a vector ranges from 5 to 15 and `multiply = 1.1`, the current
```

R/data_rescale.R
```diff
@@ -1,15 +1,26 @@
...
+#'   range of 10 will be expanded by the factor of 1.1, giving a new range of
+#'   11. Thus, the rescaled vector would range from 4.5 to 15.5.
+#' @param add If not `NULL`, `to` is ignored and `add` will be used, giving the
+#'   amount by which the actual range of `x` should be expanded. For example,
+#'   if a vector range from 5 to 15 and `add = 1`, the current range of 10 will
```
etiennebacher:
```suggestion
#'   if a vector ranges from 5 to 15 and `add = 1`, the current range of 10 will
```

R/data_rescale.R
```diff
@@ -1,15 +1,26 @@
...
+#'   will have after rescaling. To reverse-score a variable, the range should
+#'   be given with the maximum value first. See examples.
+#' @param multiply If not `NULL`, `to` is ignored and `multiply` will be used,
+#'   giving the factor by which the actual range of `x` should be expanded.
+#'   For example, if a vector range from 5 to 15 and `multiply = 1.1`, the current
+#'   range of 10 will be expanded by the factor of 1.1, giving a new range of
+#'   11. Thus, the rescaled vector would range from 4.5 to 15.5.
+#' @param add A vector of length 1 or 2. If not `NULL`, `to` is ignored and `add`
+#'   will be used, giving the amount by which the minimum and maximum of the
+#'   actual range of `x` should be expanded. For example, if a vector range from
```
etiennebacher, line 20:
```suggestion
#'   actual range of `x` should be expanded. For example, if a vector ranges from
```

etiennebacher (approved):
Thanks, minor things

etiennebacher:
> Not sure, though, if add should add (subtract) the value to each end of the range?

Hard to say, I don't have a use case for this function in mind, I guess you have one?

strengejacke:
> > Not sure, though, if add should add (subtract) the value to each end of the range?

> Hard to say, I don't have a use case for this function in mind, I guess you have one?

It seems that expected behaviour in this case would be to add `add` to min/max (I have changed this here already), see https://bsky.app/profile/tjmahr.com/post/3ke3nc3hix32u


> > we could add an alias expand_range()

> wouldn't this create a conflict with scales::expand_range() (which is called a lot by ggplot2 IIRC)?

True. So no alias.

etiennebacher:
I'm not on bluesky, can you share a screenshot? (sorry I edited your post instead of writing a comment)

strengejacke:
Oh, I thought it would be visible for non-registered, too:
image

If you like, I can send you an invitation code.

etiennebacher:
> If you like, I can send you an invitation code.

I'm fine on Mastodon for now, thanks ;)

strengejacke:
Social networks have become less important anyway, or have pretty much fallen asleep since Musk took over Twitter.

etiennebacher:
> Social networks have become less important anyway, or have pretty much fallen asleep since Musk took over Twitter.

I don't know, I just can't understand the hype for bluesky, I don't see any reason it will end up differently than twitter, but anyway that's not the topic 😄

bwiernik:
Personally I like Threads a lot among the various options

PR #478, opened 2024-02-06
data_modify() gains .it and .at

strengejacke:
``` r
library(datawizard)
data(iris)
# modify at specific positions or if condition is met
d <- iris[1:5, ]
data_modify(d, .at = "Species", .modify = as.numeric)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1          5.1         3.5          1.4         0.2       1
#> 2          4.9         3.0          1.4         0.2       1
#> 3          4.7         3.2          1.3         0.2       1
#> 4          4.6         3.1          1.5         0.2       1
#> 5          5.0         3.6          1.4         0.2       1
data_modify(d, .if = is.factor, .modify = as.numeric)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1          5.1         3.5          1.4         0.2       1
#> 2          4.9         3.0          1.4         0.2       1
#> 3          4.7         3.2          1.3         0.2       1
#> 4          4.6         3.1          1.5         0.2       1
#> 5          5.0         3.6          1.4         0.2       1

# can be combined with dots
data_modify(d, new_length = Petal.Length * 2, .at = "Species", .modify = as.numeric)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species new_length
#> 1          5.1         3.5          1.4         0.2       1        2.8
#> 2          4.9         3.0          1.4         0.2       1        2.8
#> 3          4.7         3.2          1.3         0.2       1        2.6
#> 4          4.6         3.1          1.5         0.2       1        3.0
#> 5          5.0         3.6          1.4         0.2       1        2.8
```

<sup>Created on 2024-02-06 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -28,11 +28,12 @@
 #'
 #' @param .at A character vector of variable names that should be modified. This
 #' argument is used in combination with the `.modify` argument. Note that only one
-#' of `.at` or `.if` can be provided, but not both at the same time.
+#' of `.at` or `.if` can be provided, but not both at the same time. Newly created
+#' variables in `...` are not affected by `.at`.
 #' @param .if A function that returns `TRUE` for columns in the data frame where
 #' `.if` applies. This argument is used in combination with the `.modify` argument.
 #' Note that only one of `.at` or `.if` can be provided, but not both at the same
-#' time.
+#' time. Newly created variables in `...` are not affected by `.if`.
 #' @param .modify A function that modifies the variables defined in `.at` or `.if`.
 #' This argument is used in combination with either the `.at` or the `.if` argument.
 #' Note that the modified variable (i.e. the result from `.modify`) must be either
@@ -112,6 +113,16 @@
 #' # can be combined with dots
 #' data_modify(d, new_length = Petal.Length * 2, .at = "Species", .modify = as.numeric)
 #'
+#' # note that new variables cannot be used in `.at` or `.if` arguments
+#' # this example would throw an error
+#' \dontrun{
+#' data_modify(
+#'   d,
+#'   new_length = Petal.Length * 2,
+#'   .at = c("Species", "new_length"),
+#'   .modify = as.numeric
+#' )}
+#' 
 #' # combine "data_find()" and ".at" argument
 #' out <- data_modify(
 #'   d,
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -233,6 +233,7 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
   # we need to evaluate dots here, and pass them with "do.call" to
   # the data.frame method later...
   dots <- match.call(expand.dots = FALSE)[["..."]]
+  column_names <- colnames(data)
 
   # works only for dplyr >= 0.8.0
   grps <- attr(data, "groups", exact = TRUE)
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -341,11 +341,7 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     found <- .at[.at %in% column_names]
     if (length(found)) {
       for (i in found) {
-        result <- tryCatch(
-          .modify(data[[i]]),
-          warning = function(e) e,
-          error = function(e) e
-        )
+        result <- tryCatch(.modify(data[[i]]), warning = function(e) e, error = function(e) e)
         if (inherits(result, "error")) {
           insight::format_error(
             paste0("Error in modifying variable \"", i, "\": ", result$message),
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -27,12 +27,16 @@
 #' See also 'Examples'.
 #'
 #' @param .at A character vector of variable names that should be modified. This
-#' argument is used in combination with the `.modify` argument.
+#' argument is used in combination with the `.modify` argument. Note that only one
+#' of `.at` or `.if` can be provided, but not both at the same time.
 #' @param .if A function that returns `TRUE` for columns in the data frame where
 #' `.if` applies. This argument is used in combination with the `.modify` argument.
+#' Note that only one of `.at` or `.if` can be provided, but not both at the same
+#' time.
 #' @param .modify A function that modifies the variables defined in `.at` or `.if`.
-#' This argument is used in combination with either the `.at` or the `.if`argument.
-#' If `.modify` is not provided, `.at` and `.if` are ignored.
+#' This argument is used in combination with either the `.at` or the `.if` argument.
+#' Note that the modified variable (i.e. the result from `.modify`) must be either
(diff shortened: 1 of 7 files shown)


strengejacke (commented):
@etiennebacher what do you think? If you like this idea, I would add tests.

R/data_modify.R
```diff
@@ -26,6 +26,14 @@
 #' Note that newly created variables can be used in subsequent expressions.
 #' See also 'Examples'.
 #'
+#' @param .at A character vector of variable names that should be modified. This
+#' argument is used in combination with the `.modify` argument.
+#' @param .if A function that returns `TRUE` for columns in the data frame where
+#' `.if` applies. This argument is used in combination with the `.modify` argument.
```
etiennebacher, line 32: Should specify that those cannot be used at the same time

R/data_modify.R
```diff
@@ -262,8 +297,55 @@ data_modify.grouped_df <- function(data, ...) {
...
+  }
+  if ((!is.null(.at) || !is.null(.if)) && !is.null(.modify)) {
+    # if we have ".if" defined, specify ".at"
+    if (!is.null(.if)) {
+      .at <- column_names[vapply(data[column_names], .if, logical(1))]
+    }
+    # make sure "modify" is a function
+    if (!is.function(.modify)) {
+      insight::format_error("`.modify` must be a function.")
+    }
```
etiennebacher, line 325: You can move this check earlier, before the first `if` condition so that you don't have to check `!is.null(.modify)`

R/data_modify.R
```diff
@@ -262,8 +297,55 @@ data_modify.grouped_df <- function(data, ...) {
...
+    # check for valid defined column names
+    if (!all(.at %in% column_names)) {
+      not_found <- .at[!.at %in% column_names]
+      insight::format_alert(
```
etiennebacher, line 329: This kind of check throws an error in all other functions so it should be an error here too

R/data_modify.R
```diff
@@ -262,8 +297,55 @@ data_modify.grouped_df <- function(data, ...) {
...
+      )
+    }
+    # modify variables
+    found <- .at[.at %in% column_names]
+    if (length(found)) {
+      for (i in found) {
+        data[[i]] <- .modify(data[[i]])
+      }
+    } else {
+      insight::format_alert("No variables found in the dataset that match the `.if` or `.at` argument.")
```
etiennebacher, line 361: Should be an error

tests/testthat/test-data_modify.R
```diff
@@ -490,3 +490,39 @@ test_that("data_modify works with functions that return character vectors", {
...
+  )
+  expect_message(
+    data_modify(d, .at = c("Species", "Hi", "Test"), .modify = as.numeric),
+    regex = "Variables \"Hi\" and \"Test\""
+  )
+  expect_message(
+    data_modify(d, .at = c("Hi", "Test"), .modify = as.numeric),
+    regex = "No variables found in the dataset"
+  )
+})
```
etiennebacher, line 536: Can you add a check that newly created variables are not taken into account in the `.if` argument (this should also be documented)? For example:
```r
data_modify(d, new_length = Petal.Length * 2, .if = is.numeric, .modify = as.factor)
```
doesn't convert `new_length` to factor.

R/data_modify.R
```diff
@@ -26,6 +26,14 @@
 #' Note that newly created variables can be used in subsequent expressions.
 #' See also 'Examples'.
 #'
+#' @param .at A character vector of variable names that should be modified. This
+#' argument is used in combination with the `.modify` argument.
+#' @param .if A function that returns `TRUE` for columns in the data frame where
+#' `.if` applies. This argument is used in combination with the `.modify` argument.
+#' @param .modify A function that modifies the variables defined in `.at` or `.if`.
+#' This argument is used in combination with either the `.at` or the `.if`argument.
+#' If `.modify` is not provided, `.at` and `.if` are ignored.
```
etiennebacher, line 35: I think this case should error as it is unexpected usage of `.if` and `.at`.

This should also specify that the output of `.modify` must be either of length 1 or of same length as the input. For example this errors:
```r
d <- iris[1:5, ]
data_modify(d, .if = is.numeric, .modify = function(x) unique(x))
```

etiennebacher (changes requested):
Thanks @strengejacke, I have several comments relative to the docs and error behavior.

I'm also wondering if we should have an additional function to be used in `data_modify()` so that we can clearly make the distinction between what is passed in `...` and the other args (something similar to `across()` in `dplyr`). For example, I think that
```r
mtcars |> 
  data_modify(
    x = mpg + 2,
    y = center(drat),
    group_expr(
      .if = is.numeric,
      .modify = as.factor
    )
  )
```
is more readable than
```r
mtcars |> 
  data_modify(
    x = mpg + 2,
    y = center(drat),
    .if = is.numeric,
    .modify = as.factor
  )
```
(I just use `group_expr()` as a placeholder here, I think it should have another name). What do you think?

strengejacke:
Thanks, I addressed all your comments. Regarding `group_expr()`: I personally think it's easier to handle if we have direct arguments, instead of a nested function. I'm no fan of `across()`, and usually, `.if`/`.at` and `.modify` are used _instead_ of `...`, so in most use cases, you will use of of either, but not both. Nesting functions (i.e. having functions as arguments) is in my experience highly confusing for "casual" R users (that's why I don't teach tidyverse anymore, it's far too complex for the regular R user outside the "tidyverse" bubble - and I bet the most R users are outside this expert bubble)

etiennebacher:
> usually, .if/.at and .modify are used instead of ..., so in most use cases, you will use one of the two, but not both

Maybe, but then should we forbid the use of `...` when `.modify` is provided? I'm just trying to think if there would be some confusion arising from the fact that variables defined in `...` are not included in the `.if`/`.at` arguments. But again clearly outlining this in the function documentation could be enough.

etiennebacher (approved):

strengejacke:
Yes, I'm not decided whether to allow newly defined variables in `...` in `.at` or `.if` - we could easily do this. My thought was that if I would create a new variable in `...`, _and_ use `.at/`.if`, why would I need to include the new variable there, too? What would be the benefit of modifying that I couldn't have done directly?

But maybe we decide on this later, after some experience with this feature?

etiennebacher:
For example you could create/modify variables as usual, including some numeric variables, and then say "ok let's `round()` all numeric columns". In this case you would have to remember that you have to do that in two different `data_modify()` calls because doing this in the same `data_modify()` wouldn't affect the newly created variables. 

If we forbid the use of `...` and `.modify` in the same call then you also need two `data_modify()` calls, but at least you don't have to remember the fact that you need two calls, the error message tells you that directly.

strengejacke:
hm, makes sense. I think we should change to that default behaviour.

PR #483, opened 2024-03-03
`data_summary()` works with `bayestestR::ci()`

strengejacke:
``` r
library(datawizard)
data_summary(mtcars, ci = bayestestR::ci(mpg), by = c("am", "gear"))
#> am | gear |         95% CI
#> --------------------------
#>  0 |    3 | [10.40, 21.46]
#>  0 |    4 | [17.91, 24.28]
#>  1 |    4 | [21.00, 33.64]
#>  1 |    5 | [15.08, 29.96]
```

<sup>Created on 2024-03-03 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


strengejacke:
@easystats/core-team Do we want to merge this? @DominiqueMakowski ?
It's a special case, we may extend this to other functions in the future if necessary, but for now, it might be enough? I think Dom asked for this feature somewhere...

IndrajeetPatil:
I have no strong opinion either way.

R/data_summary.R
```diff
@@ -122,7 +122,7 @@ data_summary.data.frame <- function(x, ..., by = NULL, remove_na = FALSE) {
       # bind grouping-variables and values
       summarised_data <- cbind(s[1, by], summarised_data)
       # make sure we have proper column names
-      colnames(summarised_data) <- c(by, vapply(summarise, names, character(1)))
+      colnames(summarised_data) <- c(by, unlist(lapply(summarise, names)))
```
etiennebacher, line 125: Is this a key change for this PR or just a cosmetic one? If the latter, I'd rather keep the `vapply()` call which ensures that we get the right output type.

R/data_summary.R
```diff
@@ -214,6 +218,11 @@ print.dw_data_summary <- function(x, ...) {
   if (nrow(x) == 0) {
     cat("No matches found.\n")
   } else {
+    if (all(c("CI", "CI_low", "CI_high") %in% colnames(x))) {
+      ci <- insight::format_table(x[c("CI", "CI_low", "CI_high")], ...)
+      x$CI <- x$CI_low <- x$CI_high <- NULL
+      x <- cbind(x, ci)
+    }
```
etiennebacher, line 225: can we have a snapshot test for this?

etiennebacher (approved):
I don't know if you want to wait for other feedback from the team but this is fine for me, I just have a couple of minor comments.

R/data_summary.R
strengejacke: It's because `bayestestR::ci()` may return more than one value, which would fail with `character(1)`.

strengejacke:
Haha, meanwhile we have added this code:

```r
  # check for correct length of output - must be a single value!
  if (any(lengths(out) != 1)) {
    insight::format_error(
      paste0(
        "Each expression must return a single value. Following expression",
        ifelse(sum(lengths(out) != 1) > 1, "s", " "),
        " returned more than one value: ",
        text_concatenate(vapply(dots[lengths(out) != 1], insight::safe_deparse, character(1)), enclose = "\"")
      )
    )
  }
```

and this PR doesn't work anymore. Let me try to fix this.

Issue #265, opened 2022-09-14
Getting rid of aliases for selection functions


IndrajeetPatil:
<img width="1187" alt="Screenshot 2022-09-14 at 20 50 05" src="https://user-images.githubusercontent.com/11330453/190238783-7b0546c3-6f02-47ee-98e6-ea1a26e93fcc.png">

- `get_columns()` is confusing because, unless you read the docs, it is not clear whether the function returns column names, or a list of vectors, or a data frame. `data_select()` is much better, since we have established that `data_*()` functions always return a data frame. Keeping the `get_columns()` around for long is inviting confusion.

- The same is true for `data_find()`. All other `data_*()` functions return data frames, while this one returns an atomic vector. This messes user's mental model. We should remove it in favour of `find_columns()`.

``` r
library(datawizard)
data_find(iris, starts_with("Sepal"))
#> [1] "Sepal.Length" "Sepal.Width"
```

<sup>Created on 2022-09-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This is probably something we can do in `0.7` or later.

strengejacke:
I thought the rational was that all `data_*()` functions _only_ work with data frames (not what they return). _Returning_ a data frame also works for `center()` et al, so we focus on the input when choosing the function name. But of course, we could say that `data_*()` means input _and_ output are data frames, and thus remove `data_find()` in favor of `find_columns()`. `find_columns()` sounds more natural, and that's why we added `get_columns()`, in the tradition of _insight's_ `get_*`/`find_*` pairs.

IndrajeetPatil:
> and thus remove `data_find()` in favor of `find_columns()`

Okay, that sounds reasonable to me.

> `find_columns()` sounds more natural

Maybe, but it's also more confusing. `find_column_names()` or `extract_column_names()` might have been better.
But I can make my peace with the current naming schema.

But I still think `get_columns()` should be removed in favour of `data_select()`.

The fewer functions we have that the user needs to learn, the lower the cognitive load we are imposing on them.

Users are also likely to find it more familiar to work with.

<img width="748" alt="Screenshot 2022-09-15 at 08 22 50" src="https://user-images.githubusercontent.com/11330453/190329734-24522488-2864-4e3c-b863-3599db4e7894.png">

DominiqueMakowski:
[...]

PR #502, opened 2024-05-16
Rename `group`/`group_by` arguments into `by`

strengejacke:
https://github.com/easystats/easystats/issues/404


NEWS.md
```diff
@@ -1,5 +1,12 @@
 # datawizard 0.10.1
 
+BREAKING CHANGES
+
+* Arguments named `group` or `group_by` will be deprecated in a future release.
```
etiennebacher, line 5: Why don't we deprecate now?

R/data_partition.R
```diff
@@ -38,21 +39,27 @@
...
                            verbose = TRUE,
+                           group = NULL,
                            ...) {
   # validation checks
   data <- .coerce_to_dataframe(data)
 
+  ## TODO: deprecate later
+  if (!is.null(group)) {
+    by <- group
+  }
```
etiennebacher, line 61: There should be a deprecation warning since the docs already say this arg is deprecated

NEWS.md
strengejacke: For a soft transition in our packages. Else, submitting a package with changes would cause warnings in rev-dep checks.

R/data_partition.R
strengejacke: See my comment above. I would first submit all easystats packages to CRAN so that they work without warnings. In a later update round, we can deprecate. But I have no strong opinion on this. We can also deprecate now, meaning that once we release one package, we need to release the others as well.

etiennebacher:
It *might* be due to an `xfun` update: https://github.com/yihui/xfun/issues/85

R/data_partition.R
etiennebacher: If the other packages are updated before `datawizard` is released then they will error because the arg `by` won't exist yet in the CRAN `datawizard`. So `datawizard` will have to be updated first anyway

etiennebacher: Ok I understand

etiennebacher: At the same time, when we submit `datawizard` we could just tell the CRAN team that most of the revdepcheck warnings are in packages in the same organization and that updates are coming in the few days after

NEWS.md
```diff
@@ -1,5 +1,15 @@
...
+  Please use `by` instead. This affects following functions in *datawizard*.
```
```suggestion
  Please use `by` instead. This affects the following functions in *datawizard*.
```

etiennebacher (approved):

R/data_partition.R
strengejacke: I don't mind deprecating now. The "FIXME" comment in `data_xtabulate.R`, however, requires _insight_ to be on CRAN first, and I just submitted that package this week. But we can wait with _datawizard_ anyway?

etiennebacher: Yes no problem

strengejacke:
We have some aliases, we should think about when to remove them?

```
#' @rdname text_format
#' @export
format_text <- text_format
```

```
#' @rdname recode_values
#' @export
change_code <- recode_values
```

etiennebacher:
> We have some aliases, we should think about when to remove them?

Yes, we could also address https://github.com/easystats/datawizard/issues/265 at the same time

NEWS.md
```diff
@@ -1,5 +1,15 @@
 # datawizard 0.10.1
 
+BREAKING CHANGES
+
+* Arguments named `group` or `group_by` will be deprecated in a future release.
```
etiennebacher, line 5:
```suggestion
* Arguments named `group` or `group_by` are deprecated and will be removed
  in a future release.
```

etiennebacher:
Sorry @strengejacke I meant that we could address the text aliases you mentioned and the aliases mentioned in #265 at the same time but in another PR so that we keep this one strictly about renaming args. Sorry for the confusion :grimacing:

strengejacke:
I hope we don't need to add `@rdname` tags to the deprecated functions, but I fear, it must be... Let's see what the checks say.

strengejacke:
@etiennebacher Why do we have this error for the selection-syntax vignette now?
https://github.com/easystats/datawizard/actions/runs/9127735513/job/25098644747?pr=502

etiennebacher:
> Sorry @strengejacke I meant that we could address the text aliases you mentioned and the aliases mentioned in #265 at the same time but in another PR so that we keep this one strictly about renaming args. Sorry for the confusion 😬

@strengejacke have you seen this?

strengejacke:
> Sorry @strengejacke I meant that we could address the text aliases you mentioned and the aliases mentioned in #265 at the same time but in another PR so that we keep this one strictly about renaming args. Sorry for the confusion 😬

Yeah, there are quite some files that have been changed... 🤣 But I think it's ok.

strengejacke:
> > Sorry @strengejacke I meant that we could address the text aliases you mentioned and the aliases mentioned in #265 at the same time but in another PR so that we keep this one strictly about renaming args. Sorry for the confusion 😬
> 
> @strengejacke have you seen this?

I'm too unfamiliar with git to know how to split commits into separate PRs... 😬

etiennebacher:
I'd really like to keep those separated so that we can point to a clear PR if necessary later, I can try to split this into two PRs a bit later

strengejacke:
Actually, this PR can be merged. I'm not sure it's worth the effort splitting this PR (I just saw, in VSCode, I can add a new branch from the commit where I changed the aliases and then reset this PR to the commit before - should I do this? - but I don't think there's much benefit)

The only remaining mystery is the issue in the vignette:

```
Error in `map()`:
ℹ In index: 1.
ℹ With name: vignettes/selection_syntax.Rmd.
Caused by error:
! (converted from warning) No column names that matched the required search pattern were found.
Backtrace:
 1. my_function(iris, starts_with("Sep"))
 2. datawizard::extract_column_names(data, select = selection)
 3. insight::format_warning("No column names that matched the required search pattern were found.")
 4. insight::format_alert(..., type = "warning")

Quitting from lines 193-203 [unnamed-chunk-14] (selection_syntax.Rmd)
```

Not sure why this should suddenly fail.

etiennebacher:
> I'm not sure it's worth the effort splitting this PR (I just saw, in VSCode, I can add a new branch from the commit where I changed the aliases and then reset this PR to the commit before - should I do this? 

If you talk about #503 I think it also includes the changes for the `group_by` args so merging it first wouldn't change anything I think

> The only remaining mystery is the issue in the vignette:

I can reproduce this warning locally with this branch but not with `main` so it probably comes from the alias removing. 

I will try to split this PR this weekend if it's fine for you, or are you in a hurry for this one?

strengejacke:
No, no hurry.

etiennebacher:
@strengejacke I fixed the error in the vignette here: [b8ac8ea](https://github.com/easystats/datawizard/pull/502/commits/b8ac8eaab4121616fcf85facc9b50a18b7451618). FYI [those instructions](https://gist.github.com/loilo/930f141d9acf89e9e734ffa042acd750) were helpful to know how to move the latest N commits to a new branch (still took some manual tweaking though)

`pkgdown` workflows still fail but it's something different on which we don't have much control I think

etiennebacher:
We have three different test failures but all the rest is green (??):

* ubuntu-latest 3.6: will go away when we bump version to >= 4.0 anyway
* windows-latest 4.1: I think due to `xfun` but still super weird because only happens in this case
* ubuntu-latest 4.3: `function 'chm_factor_ldetL2' not provided by package 'Matrix'`, I don't think we can do much on that

=> I'm merging

PR #506, opened 2024-05-20
Improve docs for data_to_wide

strengejacke:
Background: https://x.com/stephenjwild/status/1792294171924967920
And I recently wanted to use `data_to_wide()` and couldn't get it work at first, took some time to figure out the logic behind reshaping to wide again. Especially, the documentation of `id_cols` was confusing.


R/data_to_wide.R
```diff
@@ -4,11 +4,18 @@
...
+#' @param by The name of the column that identifies the rows in the data by
+#' which observations are grouped and the gathered data is spread into new
+#' columns. Usually, this is a variable containing an ID for observations that
+#' have been repeatedly measured. If `NULL`, it will use all remaining columns
+#' that are not in `names_from` or `values_from` as ID columns. `by` can also
+#' be a character vector with more than one name of identifier columns. See also
+#' 'Details' and 'Examples'.
+#' @param names_from The name of the column in the original data whose values
+#' will be used for naming the new columns created in the widened data. Each
+#' unique value in this column will become the name of one of these new columns.
```
etiennebacher, line 18: Last sentence is not true when `names_prefix` is provided

R/data_to_wide.R
```diff
@@ -4,11 +4,18 @@
...
+#' @param data A data frame to convert to wide format, so that it has more
+#' columns and fewer rows post-widening than pre-widening.
+#' @param by The name of the column that identifies the rows in the data by
+#' which observations are grouped and the gathered data is spread into new
+#' columns. Usually, this is a variable containing an ID for observations that
+#' have been repeatedly measured. If `NULL`, it will use all remaining columns
+#' that are not in `names_from` or `values_from` as ID columns. `by` can also
+#' be a character vector with more than one name of identifier columns. See also
+#' 'Details' and 'Examples'.
+#' @param names_from The name of the column in the original data whose values
```
etiennebacher, line 16: There can be several columns

R/data_to_wide.R
```diff
@@ -19,17 +26,38 @@
...
+#' necessary information for `data_to_wide()` is:
+#'
+#' - The name of the column that identifies the groups or repeated measurements
+#'   (`by`).
+#' - The name of the column whose _values_ will become the new column names
+#'   (`names_from`). Since these values may not necessarily reflect appropriate
+#'   column names, you can use `names_prefix` to add a prefix to each newly
+#'   created column name.
+#' - The name of the column that contains the values (`values_from`) for the
+#'   new columns that are created by `names_from`.
```
etiennebacher, line 54: same as before about multiple columns

R/data_to_wide.R
```diff
@@ -19,17 +26,38 @@
...
+#'
+#' In other words: Repeated measurements, as indicated by `by`, that are
```
etiennebacher, line 56:
```suggestion
#' In other words: repeated measurements, as indicated by `by`, that are
```

etiennebacher (changes requested):
Thanks, those are nice docs improvements. However I disagree on the renaming of `id_cols` as `by`. I don't think it makes things any clearer but instead leads to confusion since those two functions are clearly based on `tidyr::pivot_*()` functions so departing from those only for one argument is strange.

---

Regarding the tweet, I just have access to the first one since I don't have an X account, so maybe there are more explanations, but to me those `pivot_*()` functions made pivoting/unpivoting so much easier to understand than `reshape2` or `data.table` approaches.

R/data_to_wide.R
```diff
@@ -4,11 +4,18 @@
...
 #' `tidyr::pivot_wider()`.
 #'
-#' @param data A data frame to pivot.
-#' @param id_cols The name of the column that identifies the rows. If `NULL`,
-#' it will use all the unique rows.
-#' @param names_from The name of the column that contains the levels to be
-#' used as future column names.
+#' @param data A data frame to convert to wide format, so that it has more
+#' columns and fewer rows post-widening than pre-widening.
+#' @param by The name of the column that identifies the rows in the data by
```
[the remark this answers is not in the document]
strengejacke: I personally prefer `by`, both because it's in line with renaming the other arguments, especially since `by` referes to a "grouping" variable. Furthermore, `by` makes clearer that this specified variables is used for gathering those rows that are spread as new columns.
However, since we already allow `select` (easystats-name) and `cols` (tidyverse-name), we could probably use both names here, too?

R/data_to_wide.R
strengejacke: Yes, this is already mentioned in the docs:
> `by` can also be a character vector with more than one name of identifier columns.

R/data_to_wide.R
```diff
@@ -19,17 +26,38 @@
 #' [glue specification](https://glue.tidyverse.org/index.html) that uses the
 #' `names_from` columns to create custom column names. Note that the only
 #' delimiters supported by `names_glue` are curly brackets, `{` and `}`.
-#' @param values_from The name of the column that contains the values to be used
-#' as future variable values.
+#' @param values_from The name of the column in the original data that contains
```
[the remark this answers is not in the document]
strengejacke: Not sure, I haven't written this function/feature. You may add some more details to the docs here.

strengejacke:
> but to me those `pivot_*()` functions made pivoting/unpivoting so much easier to understand

True, but I think `by` even makes it even clearer. `by` refers across our easystats-packages to a variable/column that indicates "grouping". Reshaping to wide format commonly, but not in general, identifies "repeated measurements" by an "ID". And `id_cols` suggests plural, although often only one column is selected (that was also one of the criticism on Twitter)

R/data_to_wide.R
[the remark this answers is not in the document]
etiennebacher: I move this to the main thread so that we don't have arguments split between the main thread and review comments

etiennebacher:
By @strengejacke in a review comment:

> I personally prefer by, both because it's in line with renaming the other arguments across functions into by. Especially, since by refers to a "grouping" variable, and imho, by makes clearer that this specified variables is used for gathering those rows ("groups") that are spread as new columns.
>
> However, since we already allow select (easystats-name) and cols (tidyverse-name), we could probably use both argument names here, too?

etiennebacher:
> However, since we already allow select (easystats-name) and cols (tidyverse-name), we could probably use both argument names here, too?

I think we should deprecate and then remove one of the two. Those pivot functions already have a lot of args so I suggest we avoid adding bloat with aliases.

> Thus, id_cols is not even the best choice from a grammar perspective, too ;-)

I don't think there is a best choice for this, one could use `id_col` but it would suggest this must be a single column while it's not always the case. Just for comparison with other packages:

* `base::reshape()`: idvar
* `data.table::melt()`: id.vars
* `pandas` and `polars`: index

Considering a `by` argument would be quite unique (doesn't mean that other packages always make things right, but it's quite indicative).

Curious about what others think about renaming `id_cols` to `by` @easystats/core-team

strengejacke:
We slice and reshape `by` the levels of that variable - it's still my personal favorite, and we use this argument name now aggressively consistent in our packages. I don't mind deviating from other packages, especially since users are free to use those other packages if they prefer that syntax.

I just realize that when you teach R to beginners, it's really good if you can repeat yourself: "use select for..." or "use by for...". Easier to understand and remember for newbies

strengejacke:
_bump_ @DominiqueMakowski @IndrajeetPatil @bwiernik @rempsyc @mattansb

DominiqueMakowski:
I *think* I agree with @etiennebacher; id_cols' purpose was to eventually specify an index, but i don't think it is conceptually different from what `by` is expected to do

strengejacke:
> but i don't think it is conceptually different from what `by` is expected to do

Yes, that's why I thought we could use `by`, because we use it in conceptually similar ways across other functions now.

But I'm ok with sticking to `id_cols`, though I don't see the need to use that name. The `id_cols` refers to a variable, whose "levels" is used for grouping/stratification. In every other instance we now use `by`, not here though.

strengejacke:
Btw, for `data_to_long()`, we have `select` for selecting columns, but also its alias `cols` for compatibility with `pivot_longer()`

bwiernik:
I think the conceptual similarity of `by` to `id_cols` is very tenuous. I suggest we drop it

strengejacke:
Ok, if nobody likes my suggestion (😢), let's stick to `id_cols` then.

strengejacke:
`by`:

exterminador-do

PR #512, opened 2024-06-07
Allow named character vector in `select`, to rename variables

strengejacke:
``` r
library(datawizard)
data(mtcars)

data_select(mtcars, c(neu = "mpg", alt = "cyl", hoho = "wt")) |> head()
#>                    neu alt  hoho
#> Mazda RX4         21.0   6 2.620
#> Mazda RX4 Wag     21.0   6 2.875
#> Datsun 710        22.8   4 2.320
#> Hornet 4 Drive    21.4   6 3.215
#> Hornet Sportabout 18.7   8 3.440
#> Valiant           18.1   6 3.460
```

``` r

data_select(mtcars, c(neu = "mpg", "cyl", hoho = "wt")) |> head()
#>                    neu cyl  hoho
#> Mazda RX4         21.0   6 2.620
#> Mazda RX4 Wag     21.0   6 2.875
#> Datsun 710        22.8   4 2.320
#> Hornet 4 Drive    21.4   6 3.215
#> Hornet Sportabout 18.7   8 3.440
#> Valiant           18.1   6 3.460
```

<sup>Created on 2024-06-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 ## Changes
 
 * `data_select()` can directly rename selected variables when a named vector
-  is provided in `select`, e.g. `data_select(mtcars, c(new = "mpg", old = "cyl"))`.
+  is provided in `select`, e.g. `data_select(mtcars, c(new1 = "mpg", new2 = "cyl"))`.
 
 # datawizard 0.11.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # datawizard 0.11.0.1
 
+## Changes
+
+* `data_select()` can directly rename selected variables when a named vector
+  is provided in `select`, e.g. `data_select(mtcars, c(new = "mpg", old = "cyl"))`.
+
 # datawizard 0.11.0
 
 BREAKING CHANGES
--- a/R/select_nse.R
+++ b/R/select_nse.R
@@ -89,6 +89,14 @@
       # after inclusion and exclusion, the original values in "select"
       # may have changed, so we check that we only add names of valid values
       out <- stats::setNames(out, names(new_names)[new_names %in% out])
+      # check if we have any duplicated names, and if so, give an error
+      if (anyDuplicated(names(out)) > 0) {
+        insight::format_error(paste0(
+          "Following names were duplicated after renaming: ",
+          text_concatenate(names(out)[duplicated(names(out))], enclose = "`"),
+          ". Using duplicated names is no good practice and therefore discouraged. Please provide unique names."
+        ))
+      }
     }
   }
 
--- a/R/select_nse.R
+++ b/R/select_nse.R
@@ -75,6 +75,22 @@
     out <- setdiff(out, grp_vars)
   }
 
+  # for named character vectors, we offer the service to rename the columns
+  if (typeof(expr_select) == "language") {
+    # safe evaluation of the expression, to get the named vector from "select"
+    new_names <- tryCatch(eval(expr_select), error = function(e) NULL)
+    # check if we really have a named vector
+    if (!is.null(new_names) && !is.null(names(new_names))) {
+      # if so, copy names
+      all_names <- names(new_names)
+      # if some of the element don't have a name, we set the value as name
+      names(new_names)[!nzchar(all_names)] <- new_names[!nzchar(all_names)]
+      # after inclusion and exclusion, the original values in "select"
+      # may have changed, so we check that we only add names of valid values
+      out <- stats::setNames(out, names(new_names)[new_names %in% out])
+    }
+  }
+
   out
 }
 
--- a/R/select_nse.R
+++ b/R/select_nse.R
@@ -92,7 +92,7 @@
       # check if we have any duplicated names, and if so, give an error
       if (anyDuplicated(names(out)) > 0) {
         insight::format_error(paste0(
-          "Following names were duplicated after renaming: ",
+          "Following names are duplicated after renaming: ",
           text_concatenate(names(out)[duplicated(names(out))], enclose = "`"),
           ". Using duplicated names is no good practice and therefore discouraged. Please provide unique names."
         ))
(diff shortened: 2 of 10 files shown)


NEWS.md
```diff
@@ -1,5 +1,10 @@
 # datawizard 0.11.0.1
 
+## Changes
+
+* `data_select()` can directly rename selected variables when a named vector
+  is provided in `select`, e.g. `data_select(mtcars, c(new = "mpg", old = "cyl"))`.
```
```suggestion
  is provided in `select`, e.g. `data_select(mtcars, c(new1 = "mpg", new2 = "cyl"))`.
```

R/select_nse.R
```diff
@@ -75,6 +75,22 @@
     out <- setdiff(out, grp_vars)
   }
 
+  # for named character vectors, we offer the service to rename the columns
+  if (typeof(expr_select) == "language") {
```
etiennebacher, line 79: I think we should have an additional arg in `.select_nse()` like `allow_rename = FALSE` by default and we set it to `TRUE` when we call `data_select()`. Then we can have `allow_rename && typeof(expr_select) == "language"` to avoid evaluating this branch for all the other contexts in which `.select_nse()` is called.

R/select_nse.R
```diff
@@ -75,6 +75,22 @@
...
+    # safe evaluation of the expression, to get the named vector from "select"
+    new_names <- tryCatch(eval(expr_select), error = function(e) NULL)
+    # check if we really have a named vector
+    if (!is.null(new_names) && !is.null(names(new_names))) {
+      # if so, copy names
+      all_names <- names(new_names)
+      # if some of the element don't have a name, we set the value as name
```
etiennebacher, line 86:
```suggestion
      # if some of the elements don't have a name, we set the value as name
```

tests/testthat/test-data_select.R
```diff
@@ -402,3 +402,19 @@ test_that("old solution still works", {
     c("Sepal.Length", "Sepal.Width")
   )
 })
+
+test_that("data_select renames variables on the fly", {
+  data(mtcars)
+  out <- data_select(mtcars, c(new = "mpg", old = "cyl", hoho = "wt"))
+  expect_named(out, c("new", "old", "hoho"))
+  data_select(mtcars, c(new = "mpg", "cyl", hoho = "wt"))
```
etiennebacher, line 410: Missing `out <-` in several places I think.

Also maybe you can make it more readable by removing the assignment and just having:
```r
expect_named(
  data_select(mtcars, c(new = "mpg", old = "cyl", hoho = "wt")),
  c("new", "old", "hoho")
)
...
```

etiennebacher (changes requested):

etiennebacher (changes requested):
Thanks, can you also add tests for:
- what happens when the new name already exists in the data?
- what happens if the new name is used in `exclude`?

strengejacke:
> what happens when the new name already exists in the data?

It can't already exist in the data, because only the selected variables are returned. Thus, the case would be that a new name is chosen twice or more often. I'd say this should give an error?

> what happens if the new name is used in exclude?

`select` and `exclude` are processed before we check names, so any names used in `exclude` will be ignored for now (it also wouldn't make sense to specify a value in `exclude` that matches a new name, so the selected variable will be excluded, too - at least it wouldn't make sense in interactive use).

What would you suggest? Check earlier and also exclude those variables?

etiennebacher:
> I'd say this should give an error?

Agreed, I'm fine with the current changes

> What would you suggest? Check earlier and also exclude those variables?

My second point was more about whether the following works, and it does:

``` r
library(datawizard)

data_select(head(mtcars), c(drat = "mpg"), exclude = "drat")
#>                   drat
#> Mazda RX4         21.0
#> Mazda RX4 Wag     21.0
#> Datsun 710        22.8
#> Hornet 4 Drive    21.4
#> Hornet Sportabout 18.7
#> Valiant           18.1
```
We could add a test for this situation (if you agree with the current behavior)

R/select_nse.R
```diff
@@ -75,6 +76,30 @@
...
+      all_names <- names(new_names)
+      # if some of the elements don't have a name, we set the value as name
+      names(new_names)[!nzchar(all_names)] <- new_names[!nzchar(all_names)]
+      # after inclusion and exclusion, the original values in "select"
+      # may have changed, so we check that we only add names of valid values
+      out <- stats::setNames(out, names(new_names)[new_names %in% out])
+      # check if we have any duplicated names, and if so, give an error
+      if (anyDuplicated(names(out)) > 0) {
+        insight::format_error(paste0(
+          "Following names were duplicated after renaming: ",
```
etiennebacher, line 95:
```suggestion
          "Following names are duplicated after renaming: ",
```

tests/testthat/test-data_select.R
```diff
@@ -402,3 +402,38 @@ test_that("old solution still works", {
...
+  )
+  # check that excluded variables don't cause troubles
+  expect_named(
+    data_select(mtcars, c(new = "mpg", "cyl", hoho = "wt"), exclude = "wt"),
+    c("new", "cyl")
+  )
+  # error when names are not unique
+  expect_error(
+    data_select(mtcars, c(new = "mpg", old = "cyl", new = "wt")), # nolint
+    regex = "Following names were duplicated"
```
etiennebacher, line 433:
```suggestion
    regex = "Following names are duplicated"
```

etiennebacher (approved):
Apart from adding a test as mentioned in my comment above, this looks good to me, thanks!

tests/testthat/test-data_select.R
```diff
@@ -402,3 +402,38 @@ test_that("old solution still works", {
...
+  )
+  expect_error(
+    data_select(mtcars, c(new = "mpg", "cyl", cyl = "wt")), # nolint
+    regex = "Following names were duplicated"
```
etiennebacher, line 437:
```suggestion
    regex = "Following names are duplicated"
```

PR #513, opened 2024-06-09
Failing CRAN checks for recent submission (0.11.0)

### LINKED ISSUE #511: Failing CRAN checks for recent submission (0.11.0)
For old-rel: https://cran.r-project.org/web/checks/check_results_datawizard.html

```
  Errors in running code in vignettes:
  when running code in ‘tidyverse_translation.Rmd’
    ...
  +     FUN.VALUE = logical(1)))) {
  +     evaluate_chunk <- FALSE
  + }
  
  > starwars %>% data_filter(skin_color == "light", eye_color == 
  +     "brown")
  
    When sourcing ‘tidyverse_translation.R’:
  Error: could not find function "%>%"
  Execution halted
```

Not sure, why.

strengejacke:
Fixes #511


strengejacke:
@etiennebacher I removed a sentence from the vignette, which seemed to be rather outdated meanwhile. And I'm not sure if this sentence now provides any useful information for readers that contributes to clarity and wouldn't increase confusion.

vignettes/tidyverse_translation.Rmd
```diff
@@ -30,7 +30,7 @@ pkgs <- c(
 # a logical that is FALSE only if deps are not installed (cf easystats/easystats#317)
 evaluate_chunk <- TRUE
 
-if (!all(vapply(pkgs, requireNamespace, quietly = TRUE, FUN.VALUE = logical(1L)))) {
+if (!all(vapply(pkgs, requireNamespace, quietly = TRUE, FUN.VALUE = logical(1L))) || getRversion() < "4.2.0") {
```
etiennebacher, line 33: This should be < 4.1 since the base pipe appeared in 4.1.0.

Also, that would solve the issue in CRAN checks but it means that any user with 3.6 or 4.0 wouldn't be able to see the vignettes, which seems unfair (since we say we support those versions).

strengejacke: I'm not sure if the users won't see it, since vignettes are compiled and available on the CRAN and GitHub websites?

etiennebacher: I don't know if users would be able to see it or not, but in any case someone on 3.6 or 4.0 wouldn't be able to only copy-paste the code, they would also need to replace the native pipe by `magrittr`'s one. I think we can wait for the moment our oldest support version is 4.1 before making this change.

etiennebacher: Can you add a note at the top of the vignette saying that users on 3.6 or 4.0 should replace the native pipe by `magrittr`'s one?

etiennebacher (approved):

etiennebacher:
Thanks!

PR #515, opened 2024-06-21
Bump required R version to 4.0

### LINKED ISSUE #500: Tests are failing on R 3.6
cf. https://github.com/easystats/datawizard/actions/runs/9044484308/job/24853189492

Either we fix these, or we bump the minimum needed R version to 4.0 across the ecosystem.

I prefer the latter.

etiennebacher:
Necessary due to recent update of `evaluate` that has this requirement.

Close #500


strengejacke:
Where do we need that package?

etiennebacher:
Let's see if we follow this: https://github.com/easystats/easystats/issues/401#issuecomment-2182664831

strengejacke:
I think this is safe to do. This means that future updates from _parameters_ will require minimum R4.0, too, when I update package dependencies. But: (foreign) packages that import either _datawizard_ or _parameters_ either also depend on R 4.0.0 or higher, or already import other packages that reqire R 4.0.0 or higher (e.g. package _dotwhisker_ imports _gtable_, wich requires R 4.0). And furthermore, it's still possible to run all these packages in R 3.6 with older package versions. So, no serious concerns here, I think.

etiennebacher:
Went back to the original discussion on the version policy, seems like the version policy is outdated because we can't test on 3.6 anymore? 
https://easystats.github.io/easystats/articles/version_policy.html

strengejacke:
I leave it to you to decide which R version should be required. It's always possible to run older package versions on R 3.6.

etiennebacher:
Alright let's go with 4.0. We don't test on lower versions and it's available since May 2020.

PR #518, opened 2024-06-22
Unexpected missing values in `data_tabulate()`

### LINKED ISSUE #514: Unexpected missing values in `data_tabulate()`
There are no missing values in this entire data frame:

``` r
anyNA(mtcars)
#> [1] FALSE
```

And yet it shows up in `data_tabulate()` output:

``` r
datawizard::data_tabulate(mtcars, cyl, gear)
#> cyl (cyl) <numeric>
#> # total N=32 valid N=32
#> 
#> Value |  N | Raw % | Valid % | Cumulative %
#> ------+----+-------+---------+-------------
#> 4     | 11 | 34.38 |   34.38 |        34.38
#> 6     |  7 | 21.88 |   21.88 |        56.25
#> 8     | 14 | 43.75 |   43.75 |       100.00
#> <NA>  |  0 |  0.00 |    <NA> |         <NA>
```

<sup>Created on 2024-06-21 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

strengejacke:
Fixes #514


strengejacke:
Docs need to be changed roo

NEWS.md
```diff
@@ -2,6 +2,11 @@
 
 BREAKING CHANGES
 
+* The argument `include_na` in `data_tabulate()` and `data_summary()` has been
+  renamed into `remove_na`. Consequently, to mimic former behaviour, `FALSE` and
+  `TRUE` need to be switched (i.e. `remove_na = TRUE` is equivalent to the former
+  `include_na 0 FALSE`).
```
IndrajeetPatil, line 8:
```suggestion
  `include_na = FALSE`).
```

R/data_summary.R
```diff
@@ -103,10 +103,10 @@ data_summary.data.frame <- function(x, ..., by = NULL, include_na = TRUE) {
     }
     # split data, add NA levels, if requested
     l <- lapply(x[by], function(i) {
-      if (include_na && anyNA(i)) {
-        addNA(i)
-      } else {
+      if (remove_na || !anyNA(i)) {
```
IndrajeetPatil, line 106: The order of these two conditions can be switched to take advantage of short-circuiting behaviour of `||` to avoid unnecessary computation.

R/data_xtabulate.R
```diff
@@ -12,30 +12,30 @@
   }
   # frequency table
   if (is.null(weights)) {
-    if (include_na) {
-      x_table <- tryCatch(table(addNA(x), addNA(by)), error = function(e) NULL)
-    } else {
+    if (remove_na) {
       x_table <- tryCatch(table(x, by), error = function(e) NULL)
+    } else {
+      x_table <- tryCatch(table(addNA(x), addNA(by)), error = function(e) NULL)
```
IndrajeetPatil, line 18: Why do we need `tryCatch()` here? It would be good to add a comment explaining why it's needed so that nobody in future unintentionally removes it.

IndrajeetPatil (approved):
Thanks! LGTM. Few minor comments.

R/data_summary.R
strengejacke: You mean `if (!anyNA(i) ||remove_na) {`?
What would be the advantage?

R/data_xtabulate.R
strengejacke: I think because it's a `.default` method, which could accept objects that don't work with `table()`. We then later have

```r
  if (is.null(freq_table)) {
    insight::format_warning(paste0("Can't compute frequency tables for objects of class `", class(x)[1], "`."))
    return(NULL)
  }
```

R/data_summary.R
IndrajeetPatil: `remove_na` is doing something that is relevant only if `!anyNA` is `TRUE`, so in this order it's more readable because it has more logical flow considering the short-circuiting behaviour of `||`.

strengejacke: > remove_na is doing something that is relevant only if !anyNA

No, it's an OR-logical, not AND.

IndrajeetPatil: Ah, you are right. In my mind, `remove_na` was an operation and that's why I was confused.

Issue #374, opened 2023-03-09
Doc of `smoothness()` looks quite empty


etiennebacher:
The current documentation of `smoothness()` looks quite empty:

![image](https://user-images.githubusercontent.com/52219252/224015099-af6ad95c-deeb-4259-8d2d-c73c62250dd3.png)

I'm not familiar with it but is it possible to add a bit more details on the methods and some references other than a StackExchange post? For example, [`skewness()` and `kurtosis()`](https://easystats.github.io/datawizard/reference/skewness.html) are much more documented.

IndrajeetPatil:
@DominiqueMakowski _bump_

DominiqueMakowski:
Will do, thanks for the bump!

We could potentially relocate to data wizard along with other variable descriptors (albeit smoothness describes sequential series but still)

PR #519, opened 2024-06-24
[docs] document smoothness #374

DominiqueMakowski:


R/smoothness.R
```diff
@@ -12,6 +18,13 @@
 #' plot(x)
 #' smoothness(x, method = "cor")
 #' smoothness(x, method = "diff")
+#'
+#' # A boootstrapped value can also be computed
```
etiennebacher, line 22:
```suggestion
#' # A bootstrapped value can also be computed
```

R/smoothness.R
```diff
@@ -12,6 +18,13 @@
...
+#' smoothness(x, iterations = 100)
+#'
+#' # When perfectly linear, the "smoothness" is 1
+#' smoothness(1:10)
```
etiennebacher, line 26: The examples in the current docs report smoothness < 1 with "cor" and > 1 with "diff". Can you add some explanations on how to interpret those?

DominiqueMakowski: indeed, actually the "diff" method had a reversed interpretation. I fixed accordingly (but that's a breaking change)

etiennebacher: Thanks, this requires a bullet point in NEWS and fixing the tests

strengejacke: > Thanks, this requires a bullet point in NEWS and fixing the tests

image

strengejacke:
Waiting for Dom to update news or add tests...

image

R/smoothness.R
```diff
@@ -12,6 +18,17 @@
...
+#'
+#' # And closer to zero for random
+#' smoothness(rnorm(1000))
+#' smoothness(rnorm(1000), method = "diff")
+#'
 #' @return Value of smoothness.
 #' @references https://stats.stackexchange.com/questions/24607/how-to-measure-smoothness-of-a-time-series-in-r
```
etiennebacher, line 33: It would be useful to have some details on the range of the output. Based on the examples I thought it would be between 0 and 1 but one of the answers in the CrossValidated post linked says it can go to -1. What does it mean to have a smoothness of -0.5?

strengejacke: @DominiqueMakowski ?

DominiqueMakowski: 🤷

PR #528, opened 2024-07-20
Fix R CMD check errors on vignette for macOS-R 4.3

### LINKED ISSUE #527: Fix build failures on macOS 4.3
cf. https://cloud.r-project.org/web/checks/check_results_datawizard.html

```r
Version: 0.12.1
Check: running R code from vignettes
Result: ERROR 
  Errors in running code in vignettes:
  when running code in ‘tidyverse_translation.Rmd’
    ...
  > if (!all(vapply(pkgs, requireNamespace, quietly = TRUE, 
  +     FUN.VALUE = logical(1))) || getRversion() < "4.1.0") {
  +     evaluate_chunk <- FALSE
   .... [TRUNCATED] 
  
  > row("Note: In this vignette, we use the native pipe-operator, `|>`, which was introduced in R 4.1. Users of R version 3.6 or 4.0 should replace the  ..." ... [TRUNCATED] 
  
    When sourcing ‘tidyverse_translation.R’:
  Error: a matrix-like object is required as argument to 'row'
  Execution halted
  
    ‘selection_syntax.Rmd’ using ‘UTF-8’... [1s/1s] OK
    ‘standardize_data.Rmd’ using ‘UTF-8’... [4s/5s] OK
    ‘tidyverse_translation.Rmd’ using ‘UTF-8’... failed
Flavor: [r-oldrel-macos-arm64](https://www.r-project.org/nosvn/R.check/r-oldrel-macos-arm64/datawizard-00check.html)
```

As discussed elsewhere, a solution might just be to rename the `row()` function to not collide with `base::row()`.

etiennebacher:
First of all I realized that none of the workflows ran on macOS with R 4.3, which is precisely where this failure occurs. So I temporarily deactivate all other workflows and try to reproduce on this config.

---

Edit: Everything worked fine on macOS with R 4.3. I now remove the `htmltools` dependency and use raw HTML in the vignettes.

Close #527


NEWS.md
```diff
@@ -1,3 +1,8 @@
+# datawizard (development version)
+
+* Remove `htmltools` from `Suggests` in an attempt of fixing an error in CRAN
+  checks due to failures to compile some vignettes (#528).
```
IndrajeetPatil, line 4: *build a vignette

IndrajeetPatil (approved):
Thanks!

PR #535, opened 2024-08-27
Allow `n()` in `data_modify()`

strengejacke:
@etiennebacher Maybe, if checks pass, you can merge this PR for the planned release? Just a minor code change.


NEWS.md
```diff
@@ -5,6 +5,9 @@ CHANGES
 * `demean()` (and `degroup()`) now also work for nested designs, if argument
   `nested = TRUE` and  `by` specifies more than one variable.
 
+* `data_modify()` now also accepts the short-cut `1:n()` and similar, for example
+  to create an index for data groups (in grouped data frames).
```
etiennebacher, line 9:
```suggestion
* `data_modify()` now recognizes `n()`, for example to create an index for data groups 
  with `1:n()` (#535).
```

R/data_modify.R
```diff
@@ -22,6 +22,9 @@
 #'   character vector is provided, you may not add further elements to `...`.
 #' - Using `NULL` as right-hand side removes a variable from the data frame.
 #'   Example: `Petal.Width = NULL`.
+#' - For (grouped) data frames, the function `n()` can be used to count the
```
etiennebacher, line 25:
```suggestion
#' - For data frames (including grouped ones), the function `n()` can be used to count the
```

R/data_modify.R
```diff
@@ -352,8 +356,12 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify =
   # finally, we can evaluate expression and get values for new variables
   symbol_string <- insight::safe_deparse(symbol)
   if (!is.null(symbol_string) && all(symbol_string == "n()")) {
-    # "special" functions
+    # "special" functions - using "n()" just returns number of rows
     new_variable <- nrow(data)
+  } else if (!is.null(symbol_string) && length(symbol_string) == 1 && grepl("n()", symbol_string, fixed = TRUE)) {
```
etiennebacher, line 361: This feels very hacky and it seems it would break if someone passes `1:fun()` since `n()` will be detected. I also think it would open the door to more "special" functions (`first()`, `last()`, `nth()`, etc.), which can be hard to maintain in the long run.

Random thought: would parsing the expression with getParseData() and using its results make this more robust?

tests/testthat/test-data_modify.R
```diff
@@ -492,6 +492,18 @@ test_that("data_modify works with functions that return character vectors", {
 })
 
 
+test_that("data_modify 1:n() and similar works in grouped data frames", {
```
etiennebacher, line 495: Can you also check how it behaves on 

1. ungrouped dataframe
2. dataframe with 0 rows
3. maybe also inside a function to check that the `eval()` call runs in the right environment?
4. calls that are close but not identical to `n()`, e.g. `fun()`, `1:fun()`, etc.

etiennebacher (changes requested):
Thanks! I think there are cases where this implementation would fail, e.g. if someone passes `1:fun()`. Can you check for this?

You have more usecases than me with `datawizard` but it seems that we gradually try to replace all functions from `dplyr`, which I thought was not the main objective of `datawizard`. I don't know if this needed in another package internally but if not, do we actually need to implement this instead of using `dplyr`?

---

Also, if you want to add more bug fixes/features, I don't plan on releasing before the end of the week.

R/data_modify.R
strengejacke: ok, should no be catching only "n()".

> I also think it would open the door to more "special" functions 

True, but we already have `n()` in `data_summary()`, and I think it's a more common use case than first/last. Thus, this is more a "feature completion" than opening Pandorra's box ;-)

What do you mean by `getParseData()`?

tests/testthat/test-data_modify.R
strengejacke: All checks added, also behaviour for zero-row data frames.
But I have no good idea for a use-case for `fun()`?

strengejacke:
> but if not, do we actually need to implement this 

You're right, but we had this already for `data_summary()`, because it's an more often used function, and I thought it should then also work for `data_modify()`. That's the main reason why I added this. And actually, I don't use dplyr and tidyr anymore, since 99.5% of my tasks can be done with datawizard. :-)

R/data_modify.R
```diff
@@ -145,6 +149,11 @@ data_modify.default <- function(data, ...) {
 data_modify.data.frame <- function(data, ..., .if = NULL, .at = NULL, .modify = NULL) {
   dots <- eval(substitute(alist(...)))
 
+  # error for data frames with no rows...
+  if (nrow(data) == 0) {
+    insight::format_error("`data` is an empty data frame. `data_modify()` only works for data frames with at least one row.") # nolint
+  }
```
etiennebacher, line 155: Should we error here? I can imagine that this can be used in functions where the input data is not always of the right format and can have 0 rows. I think a better output would be to return a 0-row dataframe with the new columns, like:

``` r
dplyr::mutate(data.frame(x = character(0)), y = 1)
#> [1] x y
```

My comment was more to ensure that the specific `n()` works as expected when there are 0 rows.

etiennebacher (changes requested):
>  But I have no good idea for a use-case for fun()?

I think you can just check that the error message of `data_modify(mtcars, b = fun())` is comprehensible.

But before that, please check my other comment.

etiennebacher:
> You're right, but we had this already for data_summary(), because it's an more often used function, and I thought it should then also work for data_modify().

I'm also not a fan of having this in `data_summary()`. I think it adds confusion for a `tidyverse` user that could think "why is `n()` supported but not `first()` or `last()`?" and basically we don't have a good answer to that. To me the `tidyverse` and `datawizard` are more complements than substitutes: `datawizard` provides nice statistical functions that can be included in a pipeline, and it also provides dependency-free functions that we need internally in other packages anyway. Adding support for `n()` doesn't seem to meet any of those cases.

R/data_modify.R
strengejacke: Would you prefer just returning an empty data frame then, or give a message/warning and return an empty data frame?

strengejacke:
> > But I have no good idea for a use-case for fun()?
> 
> I think you can just check that the error message of `data_modify(mtcars, b = fun())` is comprehensible.
> 
> But before that, please check my other comment.

But does `fun()` necessarily result in an error? I'm not sure whether there are use cases of user-defined functions?

strengejacke:
> datawizard provides nice statistical functions that can be included in a pipeline, and it also provides dependency-free functions that we need internally in other packages anyway. Adding support for n() doesn't seem to meet any of those cases.

Yes, that was certainly the initial idea of the package, but a) there are probably other people who use datawizard _instead_ of dplyr/tidyr (e.g., @DominiqueMakowski - at least in the past - never really liked the NSE of tidyverse and prefers working with strings to identify variables etc., as do I) and b) maybe there will be other packages that prefer using datawizard for its minimal dependencies in their packages, and then it could be nice to have some features that are also supported by other packages. I personally don't mind implementing "common" features, but what do the other @easystats/maintainers think? Should we have a "stopping rule" of implementing features?

R/data_modify.R
etiennebacher: Actually, the behavior of `base` is to error, while `dplyr` and `data.table` silently return a 0-row dataframe. I think we can follow the base behavior and error here, so forget this comment.

``` r
library(data.table)

df <- data.frame(x = character(0))
dt <- data.table(x = character(0))
tib <- tibble(x = character(0))

df[["y"]] <- sum(1:3)
#> Error in `[[<-.data.frame`(`*tmp*`, "y", value = 6L): replacement has 1 row, data has 0
dt[, y := sum(1:3)][]
#> Empty data.table (0 rows and 2 cols): x,y
tib |> mutate(y = sum(1:3))
#> # A tibble: 0 × 2
#> # ℹ 2 variables: x <chr>, y <int>
```

etiennebacher:
> But does fun() necessarily result in an error? I'm not sure whether there are use cases of user-defined functions?

I just meant that we should have a test for the error below:

``` r
library(datawizard)

head(mtcars) |> 
  data_modify(x = n())
#>                    mpg cyl disp  hp drat    wt  qsec vs am gear carb x
#> Mazda RX4         21.0   6  160 110 3.90 2.620 16.46  0  1    4    4 6
#> Mazda RX4 Wag     21.0   6  160 110 3.90 2.875 17.02  0  1    4    4 6
#> Datsun 710        22.8   4  108  93 3.85 2.320 18.61  1  1    4    1 6
#> Hornet 4 Drive    21.4   6  258 110 3.08 3.215 19.44  1  0    3    1 6
#> Hornet Sportabout 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2 6
#> Valiant           18.1   6  225 105 2.76 3.460 20.22  1  0    3    1 6

head(mtcars) |> 
  data_modify(x = fun())
#> Error: There was an error in the first expression. Attempt to apply
#>   non-function. Possibly misspelled or not yet defined?
```

strengejacke:
@easystats/core-team Any opinions to
https://github.com/easystats/datawizard/pull/535#issuecomment-2316072611
https://github.com/easystats/datawizard/pull/535#issuecomment-2316293369

?

rempsyc:
> I personally don't mind implementing "common" features, but what do the other easystats/maintainers think?

Personally, I think the harm in adding extra features is limited (although I can also see Etienne's point about creating confusion). I think the potential benefit of new features ouweight the cons (although it also adds to future maintenance and development), so as long as you're motivated to do the associated work Daniel (and I think you are), I'd say go ahead 🥸

etiennebacher:
So do we put this in 0.13.0?

strengejacke:
> @easystats/core-team Any opinions to [#535 (comment)](https://github.com/easystats/datawizard/pull/535#issuecomment-2316072611) [#535 (comment)](https://github.com/easystats/datawizard/pull/535#issuecomment-2316293369)
> 
> ?

@easystats/maintainers any further comments on this? :-)

rempsyc:
Add it!

etiennebacher (approved):
Seems like I'm the only one blocking this but I'm not a big `datawizard` user compared to you so feel free to merge this once the bullet point in NEWS has been put in the good section

strengejacke:
A package maintainer not using the packages (s)he maintains, that's also some kind of punk-rock! 😄 
beavis-and-butthead-headbanging

etiennebacher:
I had some usecases for it when I started contributing but I'm full in tidyverse code for all my projects now and rarely rely on `datawizard` (mostly using the summary / codebook functions) :sweat_smile:

PR #538, opened 2024-09-08
data_rename() works with named vector

strengejacke:


tests/testthat/test-data_rename.R
```diff
@@ -14,6 +14,10 @@ test_that("data_rename works with one or several replacements", {
     ),
     c("length", "width", "Petal.Length", "Petal.Width", "Species")
   )
+  expect_named(
+    data_rename(test, c(length = "Sepal.Length", width = "Sepal.Width")),
+    c("length", "width", "Petal.Length", "Petal.Width", "Species")
+  )
```
etiennebacher, line 20: Looks good to me, can you add a test about what happens when someone passes a partially named vector, e.g. `data_rename(test, c(length = "Sepal.Length", "Sepal.Width"))`? I suppose it would error but I just want to ensure it throws a nice error message.

Same thing for the case where the user passes a named vector and a replacement, e.g. `data_rename(test, c(length = "Sepal.Length"), replacement = "Sepal.Width")`

strengejacke: First example will remove column name, second will ignore `replacement`. First example is in line with the current behaviour: 
`data_rename(iris, c("Sepal.Length", "Sepal.Width"), replacement = c("aaa", ""))`

For the second case, I have made this more explicit in the docs.

strengejacke: See https://github.com/easystats/datawizard/pull/539/

PR #539, opened 2024-09-09
`data_rename()` doesn't accept empty replacement anymore

strengejacke:


R/data_rename.R
```diff
@@ -45,6 +47,14 @@
 #' # Change all
 #' head(data_rename(iris, replacement = paste0("Var", 1:5)))
 #'
+#' # remove one column name
+#' head(data_rename(
+#'   iris,
+#'   pattern = c("Sepal.Length", "Sepal.Width"),
+#'   replacement = c("keep_length", "")
+#' ))
```
etiennebacher, line 55: I don't think this is a desirable behavior, the function to rename shouldn't deal with variable selection. I think this is prone to bugs and it would probably be better to error in this case.

Worth noting that `dplyr` errors and `data.table` just gives an empty name but none of them remove the column:

``` r
dplyr::rename(mtcars, `` = mpg)
# Error: attempt to use zero-length variable name

foo <- data.table::as.data.table(mtcars)
data.table::setnames(foo, c("mpg", "drat"), c("foo", ""))
head(foo)
#>      foo   cyl  disp    hp          wt  qsec    vs    am  gear  carb
#>    <num> <num> <num> <num> <num> <num> <num> <num> <num> <num> <num>
#> 1:  21.0     6   160   110  3.90 2.620 16.46     0     1     4     4
#> 2:  21.0     6   160   110  3.90 2.875 17.02     0     1     4     4
#> 3:  22.8     4   108    93  3.85 2.320 18.61     1     1     4     1
#> 4:  21.4     6   258   110  3.08 3.215 19.44     1     0     3     1
#> 5:  18.7     8   360   175  3.15 3.440 17.02     0     0     3     2
#> 6:  18.1     6   225   105  2.76 3.460 20.22     1     0     3     1
```

etiennebacher (changes requested):

R/data_rename.R
etiennebacher: And no matter the decision, I think we should have tests for this (maybe this one is already tested) and the case where there is a named vector and `replacement` is ignored :-)

strengejacke: Sorry, doesn't remove columns, but column _names_.
But I don't mind throwing an error. Note that formerly removing column names was possible.

etiennebacher: Oh, ok I misunderstood then. Throwing an error sounds better, I can't think of a case where would someone want an empty name

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # datawizard (development)
 
+BREAKING CHANGES
+
+* `data_rename()` now errors when the `replacement` argument contains `NA` values
+  or empty strings.
```
```suggestion
  or empty strings (#539).
```

R/data_rename.R
```diff
@@ -82,6 +82,21 @@ data_rename <- function(data,
...
   }
 
+  # coerce to character
+  replacement <- as.character(replacement)
+
+  # check if `replacement` has no empty strings and no NA values
+  invalid_replacement <- is.na(replacement) | !nzchar(replacement)
+  if (any(invalid_replacement)) {
+    insight::format_error(
+      "`replacement` is not allowed to have `NA` or empty strings.",
```
etiennebacher, line 92: Sorry for being nitpicky but this error is a bit confusing if someone passes `pattern = c(test = "Species", "Sepal.Length")` (we didn't provide `replacement` so it shouldn't appear in the message). When we check that `pattern` has names above, can we throw an error like "Either name all elements of `pattern` or use `replacement`" if some elements don't have any names?

etiennebacher (commented):
Thank you, LGTM except for one error message to clarify but apart from that feel free to merge

etiennebacher (approved):

PR #545, opened 2024-10-01
Remove Twitter-tag from Authors@R

strengejacke:
See discussion in https://github.com/easystats/insight/pull/936.

This NOTE was already on win-builder, but I didn't realize. I now uploaded _insight_ to win-builder again, removing Twitter IDs. If this resolves the NOTE, we can merge this PR (it's likely to cause the same problems with _datawizard_ then).


strengejacke:
Win-builder checks now pass after removing Twitter handles. We should merge this PR.

PR #546, opened 2024-10-02
Release 0.13.0

etiennebacher:
Close #543

PR #547, opened 2024-10-02
Default values of `to_numeric()`

### LINKED ISSUE #544: Default values of `to_numeric()`
Regarding https://github.com/easystats/easystats/issues/434, I was thinking what would be a good default for the `dummy_factors` argument in `to_numeric()`? Currently, the default is `TRUE`, splitting one variable into multiple ones.

Personally, I think this - as default - is a bit odd and I would expect a "numeric" variant of the factor, what `dummy_factors = FALSE` returns by default.

WDYT?

strengejacke:
Fixes #544

### DIFF
--- a/R/to_numeric.R
+++ b/R/to_numeric.R
@@ -17,11 +17,11 @@
 #' @inheritParams extract_column_names
 #' @inheritParams categorize
 #'
-#' @note By default, `to_numeric()` converts factors into "binary" dummies, i.e.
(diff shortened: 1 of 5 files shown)


etiennebacher (approved):
Thanks, can you just add a bullet point in news before merging?

strengejacke:
> Thanks, can you just add a bullet point in news before merging?

I would add this to #546

PR #551, opened 2024-10-10
Allow : in select strings

strengejacke:


R/select_nse.R
```diff
@@ -146,31 +146,63 @@
   # use colnames because names() doesn't work for matrices
   columns <- colnames(data)
   if (isTRUE(regex)) {
+    # string is a regular expression
     grep(x, columns)
   } else if (length(x) == 1L && x == "all") {
+    # string is "all" - select all columns
     seq_along(data)
+  } else if (any(grepl(":", x, fixed = TRUE))) {
+    # special pattern, as string (e.g.select = c("cyl:hp", "am")). However,
```
etiennebacher, line 155: Can you add an item in the "Possibilities" comment above?

tests/testthat/test-select_nse.R
```diff
@@ -138,3 +138,22 @@ test_that(".select_nse: works with function and namespace", {
   out <- fun(insight::find_predictors(model, effects = "fixed", flatten = TRUE))
   expect_identical(out, iris["Petal.Width"])
 })
+
+test_that(".select_nse: allow character vector with :", {
```
etiennebacher, line 142: Can you add a test for a case where one does `col1:col2` but `col1` comes after `col2` in the dataset?

etiennebacher (approved):
Thanks, LGTM, just a couple of comments + NEWS item to add

PR #553, opened 2024-10-10
Add `row_count()` to count specific values row-wise

strengejacke:
@etiennebacher not sure if you think `exact` is the best argument name, I'm open for any better suggestions.


etiennebacher:
I feel I'm always the one reluctant to add new functions here (sorry about that :sweat_smile:) but this feels like a very narrow usecase, especially since it's possible to do it in a few lines (except if I'm missing sth):

``` r
library(datawizard)

dat <- data.frame(
  c1 = c("1", "2", NA, "3"),
  c2 = c(NA, "2", NA, "3"),
  c3 = c(NA, 4, NA, NA),
  c4 = c(2, 3, 7, Inf)
)

data_modify(
  dat,
  count_2 = rowSums(dat == 2, na.rm = TRUE),
  count_3 = rowSums(dat == 3, na.rm = TRUE)
)
#>     c1   c2 c3  c4 count_2 count_3
#> 1    1 <NA> NA   2       1       0
#> 2    2    2  4   3       2       1
#> 3 <NA> <NA> NA   7       0       0
#> 4    3    3 NA Inf       0       2
```

I haven't looked at the other PRs on `row_` yet, but maybe it would be more efficient to allow rowwise operations rather than implementing several `row_` functions? I don't think this would be easy though.

strengejacke:
> (unless I'm missing sth)

Yes, you're missing something :-)

``` r
library(datawizard)

dat <- data.frame(
  c1 = c("1", "2", NA, "3"),
  c2 = c(NA, "2", NA, "3"),
  c3 = c(NA, 4, NA, NA),
  c4 = c(2, 3, 7, Inf)
)

data_modify(
  dat,
  count__2 = rowSums(dat == 2, na.rm = TRUE),
  count_string_2 = rowSums(dat == "2", na.rm = TRUE)
)
#>     c1   c2 c3  c4 count__2 count_string_2
#> 1    1 <NA> NA   2        1              1
#> 2    2    2  4   3        2              2
#> 3 <NA> <NA> NA   7        0              0
#> 4    3    3 NA Inf        0              0

row_count(dat, count = 2)
#> [1] 1 2 0 0
row_count(dat, count = "2")
#> [1] 1 2 0 0
row_count(dat, count = 2, exact = TRUE)
#> [1] 1 0 0 0
row_count(dat, count = "2", exact = TRUE)
#> [1] 0 2 0 0
```

NEWS.md
```diff
@@ -6,6 +6,8 @@ CHANGES
   variables, can now also be a character vector with quoted variable names,
   including a colon to indicate a range of several variables (e.g. `"cyl:gear"`).
 
+* New function `row_count()`, to calculate row-wise sums of specific values.
```
etiennebacher, line 9:
```suggestion
* New function `row_count()`, to count specific values row-wise.
```

R/row_count.R
```diff
@@ -0,0 +1,101 @@
+#' @title Row means or sums (optionally with minimum amount of valid values)
```
etiennebacher, line 1: to change

R/row_count.R
```diff
@@ -0,0 +1,101 @@
...
+#' @name row_count
+#' @description `row_count()` mimics base R's `rowSums()`, with sums for a
+#' specific value indicated by `count`. Hence, it is equivalent to
+#' `rowSums(x == count, na.rm = TRUE)`, but offers some more options, including
+#' strict comparisons: Comparisons using `==` coerce values to atomic vectors,
```
etiennebacher, line 6:
```suggestion
#' strict comparisons. Comparisons using `==` coerce values to atomic vectors,
```

R/row_count.R
```diff
@@ -0,0 +1,101 @@
...
+#' thus both `2 == 2` and `"2" == 2` are `TRUE`. In `row_count()`, it is also
+#' possible to make "type safe" comparisons using the `exact` argument, where
+#' `"2" == 2` is not treated as identical.
```
etiennebacher, line 9:
```suggestion
#' `"2" == 2` is not true.
```

R/row_count.R
```diff
@@ -0,0 +1,101 @@
...
+#'
+#' @param data A data frame with at least two columns, where number of specific
+#' values are counted row-wise.
+#' @param count The value for which the row sum should be computed. May be a
+#' numeric value, a character string (for factors or character vectors), `NA` or
+#' `Inf`.
+#' @param exact Logical, if `TRUE`, `count` matches only values of same type
```
etiennebacher, line 16:
```suggestion
#' @param exact Logical. If `TRUE`, `count` matches only values of same type
```

R/row_count.R
```diff
@@ -0,0 +1,101 @@
...
+#' row_count(dat, count = NA)
+#'
+#' dat <- data.frame(
+#'   c1 = c("1", "2", NA, "3"),
+#'   c2 = c(NA, "2", NA, "3"),
+#'   c3 = c(NA, 4, NA, NA),
+#'   c4 = c(2, 3, 7, Inf)
+#' )
+#' # count all 2s and "2"s per row
+#' row_count(dat, count = 2)
```
etiennebacher, line 46: this one already appears above

R/row_count.R
```diff
@@ -0,0 +1,101 @@
...
+    # to achieve this, we simply compute the comparison on numeric or non-numeric
+    # columns only
+    if (isTRUE(exact)) {
+      numeric_columns <- vapply(data, is.numeric, TRUE)
+      if (is.numeric(count)) {
+        data <- data[numeric_columns]
+      } else {
+        data <- data[!numeric_columns]
+      }
+    }
```
etiennebacher, line 98: This doesn't take into account other coercions, such as:

```r
as.Date("2020-01-01") == "2020-01-01"
#> TRUE
```
Maybe the value of `count` gives the type to look for and then we only retain columns that match this type?

tests/testthat/test-row_count.R
```diff
@@ -0,0 +1,39 @@
+test_that("row_count", {
+  d_mn <- data.frame(
+    c1 = c(1, 2, NA, 4),
+    c2 = c(NA, 2, NA, 5),
+    c3 = c(NA, 4, NA, NA),
+    c4 = c(2, 3, 7, 8)
+  )
+  expect_identical(row_count(d_mn, count = 2), c(1, 2, 0, 0))
+  expect_identical(row_count(d_mn, count = NA), c(2, 0, 3, 1))
```
etiennebacher, line 9: can you add a test with `count = NULL` and one with a 0-row dataframe (just to check we don't accidentally create a row in this case)?

etiennebacher (commented):
Fair enough, I should have read the changes before commenting ;)

A few comments

strengejacke:
Great! What about the argument name `exact`? (See my initial comment)
I thought an alternative could be `type_safe`?

etiennebacher:
I think `allow_coercion = TRUE/FALSE` would be clearer

R/row_count.R
strengejacke: I changed the above example, because I also want it here for the contrast.

tests/testthat/test-row_count.R
strengejacke: count = NULL is in line 23 (`count` is checked first, before values of `select`)

R/row_count.R
strengejacke: What would you suggest how to solve this?

```r
typeof(as.Date("2020-01-01"))
#> [1] "double"
```

The type of Date variables is double, but it compares to character vector (like in your example). Using the `class()` instead is difficult for factor-comparison, because `count = "a"` should also match factor levels `a` (else, it's hard to define that value), so we can't compare classes, too.

comparisons of types is difficult, comparisons of class is difficult.

etiennebacher: But if we're in the strict case, should we count both a character "a" and a factor level "a"?

strengejacke: How would you count factor levels then?

strengejacke: (if you only want to count specific factor level, how would you define the `count` argument?)

etiennebacher: Maybe an extra arg `use_levels`? I don't have a perfect solution for this for now, but my point is that if we have  `allow_coercion = FALSE` then it should prevent all coercion, not only numeric <-> non-numeric.

strengejacke: An extra argument is probably too much. But I just thought that `count = factor("a")` may work?

etiennebacher: Yes, sounds good to me

R/row_count.R
```diff
@@ -0,0 +1,124 @@
...
+  } else {
+    # comparisons in R using == coerce values into a atomic vector, i.e.
+    # 2 == "2" is TRUE. If `allow_coercion = FALSE`, we only want 2 == 2 or
+    # "2" == "2" (i.e. we want exact types to be compared only)
+    if (isFALSE(allow_coercion)) {
+      # we need the "type" of the count-value - we use class() instead of typeof(),
+      # because the latter sometimes returns unsuitable classes/types. compare
+      # typeof(as.Date("2020-01-01")), which returns "double".
+      count_type <- class(count)[1]
+      valid_columns <- vapply(data, function(i) identical(class(i)[1], count_type), TRUE)
```
etiennebacher, line 111: I think `inherits()` is more robust here

etiennebacher:
Thanks!

PR #561, opened 2024-11-11
Suggestion of new function: `describe_missing()`

### LINKED ISSUE #454: Suggestion of new function: `describe_missing()`
When writing (psychology) scientific papers, great care must be taken in reporting the state of item-level missing data for each psychological questionnaire. For example, Parent (2013) writes:

> I recommend that authors (a) state their tolerance level for missing data by scale or subscale (e.g., “We calculated means for all subscales on which participants gave at least 75% complete data”) and then (b) report the individual missingness rates by scale per data point (i.e., the number of missing values out of all data points on that scale for all participants) and the maximum by participant (e.g., “For Attachment Anxiety, a total of 4 missing data points out of 100 were observed, with no participant missing more than a single data point”).

In order to comply with this recommandation, I have developed the function `nice_na()`, which nicely summarizes `NA` values according to those guidelines. The function describes both absolute and percentage values of specified column lists and supports specifying scales through regex. Reprex:

``` r
library(rempsyc)

# If the questionnaire items start with the same name, e.g.,
set.seed(15)
fun <- function() {
  c(sample(c(NA, 1:10), replace = TRUE), NA, NA, NA)
}
df <- data.frame(
  ID = c("idz", NA),
  open_1 = fun(), open_2 = fun(), open_3 = fun(),
  extrovert_1 = fun(), extrovert_2 = fun(), extrovert_3 = fun(),
  agreeable_1 = fun(), agreeable_2 = fun(), agreeable_3 = fun()
)

head(df, 3)
#>     ID open_1 open_2 open_3 extrovert_1 extrovert_2 extrovert_3 agreeable_1
#> 1  idz      4     NA      1           5           6           1           7
#> 2 <NA>      9      4      3           1          10          NA           7
#> 3  idz      1      4      1           9           2          NA           8
#>   agreeable_2 agreeable_3
#> 1           7           9
#> 2           7           2
#> 3           7           8

# One can list the scale names directly:
nice_na(df, scales = c("ID", "open", "extrovert", "agreeable"))
#>                       var items na cells na_percent na_max na_max_percent
#> 1                   ID:ID     1  7    14      50.00      1            100
#> 2           open_1:open_3     3 11    42      26.19      3            100
#> 3 extrovert_1:extrovert_3     3 17    42      40.48      3            100
#> 4 agreeable_1:agreeable_3     3 10    42      23.81      3            100
#> 5                   Total    10 45   140      32.14     10            100
#>   all_na
#> 1      7
#> 2      3
#> 3      3
#> 4      3
#> 5      2
```

<sup>Created on 2023-09-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## Would you like this function to migrate from `rempsyc` to `datawizard`?

For the name, I was thinking `data_missing_items` or just `data_missing` since it also works without scale items and it is similar to our other `data_` functions like `data_duplicated`. It could also be `describe_missing` in line with `describe_distribution`.

rempsyc:
Fixes #454


R/describe_missing.R
```diff
@@ -0,0 +1,115 @@
...
+#' reporting item-level missingness per scale, as well as a participant's maximum
+#' number of missing items by scale. For example, Parent (2013) writes:
+#'
+#' *I recommend that authors (a) state their tolerance level for missing data by scale
+#' or subscale (e.g., "We calculated means for all subscales on which participants gave
+#' at least 75% complete data") and then (b) report the individual missingness rates
+#' by scale per data point (i.e., the number of missing values out of all data points
+#' on that scale for all participants) and the maximum by participant (e.g., "For Attachment
+#' Anxiety, a total of 4 missing data points out of 100 were observed, with no participant
+#' missing more than a single data point").*
```
etiennebacher, line 15: This sounds a bit too much focused on survey data while this function can be interesting for all kinds of data. I'd rather keep the first or two first sentences here and move the rest in a specific section in 'Details' (but even there, this seems very field-specific).

R/describe_missing.R
```diff
@@ -0,0 +1,115 @@
...
+#'
+#' @param data The data frame to be analyzed.
+#' @param vars Variable (or lists of variables) to check for missing values (NAs).
```
etiennebacher, line 18: We use `select`, `exclude`, etc. in all other dataframe functions, I think we should here as well.

etiennebacher (changes requested):
Thank you, I think it would be good to have `describe_missing()` but the way it is implemented and documented looks very field-specific to me. I find the output of `skimr::skim()` easier to understand with `n_missing` and `complete_rate` for instance. I'm also not familiar at all with aggregating stats on missing values across several variables (e.g. `Ozone:Wind`) and the default output looks unexpected to me (I'd rather expect one row per variable).

R/describe_missing.R
```diff
@@ -0,0 +1,115 @@
...
+#' @param scales The scale names to check for missing values (as a character vector).
+#' @keywords missing values NA guidelines
+#' @return A dataframe with the following columns:
+#'  - `var`: Variables selected.
+#'  - `items`: Number of items for selected variables.
```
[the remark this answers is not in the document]
rempsyc: Hum, so in this case "number of items" refers to the number of columns selected for each "scale" or combination of variables. Maybe I should use that instead, as I'm afraid `unique_values` would suggest unique responses for a given column.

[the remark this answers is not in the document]
rempsyc: It is indeed specific as in psychology we tend to think of variables as made of several "items". So items 1-10 create a variable such as a personality trait "extroversion". I'm not sure how to call it because "variable" might be confused with "scale" (i.e., a composite score). Maybe I could just rename that output column "columns", but I'm open to your suggestions if you have more.

R/describe_missing.R
```diff
@@ -0,0 +1,115 @@
...
+#' at least 75% complete data") and then (b) report the individual missingness rates
+#' by scale per data point (i.e., the number of missing values out of all data points
+#' on that scale for all participants) and the maximum by participant (e.g., "For Attachment
+#' Anxiety, a total of 4 missing data points out of 100 were observed, with no participant
+#' missing more than a single data point").*
+#'
+#' @param data The data frame to be analyzed.
+#' @param vars Variable (or lists of variables) to check for missing values (NAs).
+#' @param scales The scale names to check for missing values (as a character vector).
+#' @keywords missing values NA guidelines
```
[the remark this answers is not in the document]
rempsyc: They are in fact not recommended. I have removed them.

R/describe_missing.R
rempsyc: Here it works a little bit differently than `select` elsewhere. `vars` takes a list of list of strings (such as `list(c("Ozone", "Solar.R", "Wind"), c("Temp", "Month", "Day"))`) to take into account the nested structure of the items / columns. I can rename it to `select`, but do you think it will create confusion or expectations that it should rely on and work with `.select_nse`? Or should we include `select` and `exclude` in addition to `vars`? I'm not sure how `.select_nse` could accommodate the nested structure like I'm doing right now 🤔

R/describe_missing.R
rempsyc: I moved everything after "Some authors recommend" to `@details`.

Also, I think the way I see it, is that a lot of packages and functions can report basic missing data features, like `skimr::skim()` (that's the "easy" part). What is missing is a way to handle, as you highlight, survey data in that field-specific way. I thought it still fits with `datawizard` even if offers additional field-specific features, although we can probably try to make it more general for other users. In the details section, I added a paragraph adding more context about scales as used in psychology:
```
#' In psychology, it is common to ask participants to answer questionnaires in
#' which people answer several questions about a specific topic. For example,
#' people could answer 10 different questions about how extroverted they are.
#' In turn, researchers calculate the average for those 10 questions (called
#' items). These questionnaires are called (e.g., Likert) "scales" (such as the
#' Rosenberg Self-Esteem Scale, also known as the RSES).
```

rempsyc: I suppose one question we have to answer is: do we want to have `describe_missing` only report basic missing info that is field-general a bit more like `skim()`, OR we do we also want it to include the features specific to the survey format?

rempsyc: We can also definitely rename the column names to use `missing_` instead of `na_`

rempsyc:
Thanks for the feedback and comments! We can definitely rename the column names for more clarity e.g., to use `missing_` instead of `na_` and other suggestions. I can also add a new column `complete_rate` to mirror `skim()`. Otherwise, skim() and `describe_missing()` have the same relative structure (variables in the first column and aggregate stats on the other columns).

> the default output looks unexpected to me (I'd rather expect one row per variable).

There is one row per variable / scale, but each variable / scale can be defined by multiple items / columns. This is also totally an optional feature and without specifying scale it will also provide the same output as skim.

rempsyc:
Ok so I changed the default so that when no scale or variable are specified, all columns are reported on separate rows:

However, this behaviour is overwritten if scales or variables are specified: 

``` r
library(datawizard)

# Use the entire data frame
set.seed(15)
fun <- function() {
  c(sample(c(NA, 1:10), replace = TRUE), NA, NA, NA)
}
df <- data.frame(
  ID = c("idz", NA),
  openness_1 = fun(), openness_2 = fun(), openness_3 = fun(),
  extroversion_1 = fun(), extroversion_2 = fun(), extroversion_3 = fun(),
  agreeableness_1 = fun(), agreeableness_2 = fun(), agreeableness_3 = fun()
)
describe_missing(df)
#>           variable n_columns n_missing cells missing_percent complete_percent
#> 1               ID         1         7    14           50.00            50.00
#> 2       openness_1         1         4    14           28.57            71.43
#> 3       openness_2         1         4    14           28.57            71.43
#> 4       openness_3         1         3    14           21.43            78.57
#> 5   extroversion_1         1         6    14           42.86            57.14
#> 6   extroversion_2         1         6    14           42.86            57.14
#> 7   extroversion_3         1         5    14           35.71            64.29
#> 8  agreeableness_1         1         3    14           21.43            78.57
#> 9  agreeableness_2         1         4    14           28.57            71.43
#> 10 agreeableness_3         1         3    14           21.43            78.57
#> 11           Total        10        45   140           32.14            67.86
#>    missing_max missing_max_percent all_missing
#> 1            1                 100           7
#> 2            1                 100           4
#> 3            1                 100           4
#> 4            1                 100           3
#> 5            1                 100           6
#> 6            1                 100           6
#> 7            1                 100           5
#> 8            1                 100           3
#> 9            1                 100           4
#> 10           1                 100           3
#> 11          10                 100           2

# If the questionnaire items start with the same name,
# one can list the scale names directly:
describe_missing(df, scales = c("ID", "openness", "extroversion", "agreeableness"))
#>                          variable n_columns n_missing cells missing_percent
#> 1                              ID         1         7    14           50.00
#> 2           openness_1:openness_3         3        11    42           26.19
#> 3   extroversion_1:extroversion_3         3        17    42           40.48
#> 4 agreeableness_1:agreeableness_3         3        10    42           23.81
#> 5                           Total        10        45   140           32.14
#>   complete_percent missing_max missing_max_percent all_missing
#> 1            50.00           1                 100           7
#> 2            73.81           3                 100           3
#> 3            59.52           3                 100           3
#> 4            76.19           3                 100           3
#> 5            67.86          10                 100           2

# Otherwise you can provide nested columns manually:
describe_missing(df,
                 select = list(
                   c("ID"),
                   c("openness_1", "openness_2", "openness_3"),
                   c("extroversion_1", "extroversion_2", "extroversion_3"),
                   c("agreeableness_1", "agreeableness_2", "agreeableness_3")
                 )
)
#>                          variable n_columns n_missing cells missing_percent
#> 1                              ID         1         7    14           50.00
#> 2           openness_1:openness_3         3        11    42           26.19
#> 3   extroversion_1:extroversion_3         3        17    42           40.48
#> 4 agreeableness_1:agreeableness_3         3        10    42           23.81
#> 5                           Total        10        45   140           32.14
#>   complete_percent missing_max missing_max_percent all_missing
#> 1            50.00           1                 100           7
#> 2            73.81           3                 100           3
#> 3            59.52           3                 100           3
#> 4            76.19           3                 100           3
#> 5            67.86          10                 100           2
```

<sup>Created on 2024-12-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher:
I feel like most unresolved comments and questions regarding the documentation and the implementation are related to the scope of this function. I'd rather have a "generalist" function à la `skimr` rather than something specialized for psychology that I think could live in the `rempsyc` package.

@easystats/core-team what do you think? Are you interested in having some of those field-specific features in this function?

mattansb:
I tend to agree. This function should be more general purpose - and maybe a psych-centric wrapper can be housed in @rempsyc 's package (I also just now noticed your handle is the name of the package 😅)

DominiqueMakowski:
If I understand, the main outstanding issue is what to do with the `"scales"` argument. I would indeed remove it (soz Rémi ^^) and replace it by a `by` argument as in our other function. If users want to compute the amount of missing per dimension, they should do it using a more traditional approach and first pivot to longer and then run `describe_missing(select="item", by="dimension")` otherwise I'm afraid it gets messy if we have a bespoke `scales` argument only for this function

rempsyc:
Alright, in this case, I think I can introduce `select`, `exclude`, and `by` and make it more consistent with the rest of `datawizard` 🤓

rempsyc:
Alright, this is a much simplified version which now also support "by". So this is what I have so far:

``` r
library(datawizard)

describe_missing(airquality, select = "Ozone:Temp")
#>   variable n_missing missing_percent complete_percent
#> 1    Ozone        37           24.18            75.82
#> 2  Solar.R         7            4.58            95.42
#> 3     Wind         0            0.00           100.00
#> 4     Temp         0            0.00           100.00
#> 5    Total        44            7.19            92.81

describe_missing(airquality, exclude = "Ozone:Temp")
#>   variable n_missing missing_percent complete_percent
#> 1    Month         0               0              100
#> 2      Day         0               0              100
#> 3    Total         0               0              100

# Testing the 'by' argument for survey scales
set.seed(15)
fun <- function() {
  c(sample(c(NA, 1:10), replace = TRUE), NA, NA, NA)
}
df <- data.frame(
  ID = c("idz", NA),
  openness_1 = fun(), openness_2 = fun(), openness_3 = fun(),
  extroversion_1 = fun(), extroversion_2 = fun(), extroversion_3 = fun(),
  agreeableness_1 = fun(), agreeableness_2 = fun(), agreeableness_3 = fun()
)

df_long <- reshape_longer(
  df,
  select = -1,
  names_sep = "_",
  names_to = c("dimension", "item"))

describe_missing(df_long, 
                 select = -c(1, 3), 
                 by = "dimension")
#>        variable n_missing missing_percent complete_percent
#> 1 agreeableness        10           23.81            76.19
#> 2  extroversion        17           40.48            59.52
#> 3      openness        11           26.19            73.81
#> 4         Total        38           15.08            84.92
```

<sup>Created on 2024-12-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


Anything else you'd find desirable in the function?

NEWS.md
```diff
@@ -16,6 +16,10 @@ BREAKING CHANGES AND DEPRECATIONS
   - if `select` (previously `pattern`) is a named vector, then all elements
     must be named, e.g. `c(length = "Sepal.Length", "Sepal.Width")` errors.
 
+NEW FUNCTIONS
+
+* `describe_missing()`, to report on missing values in a data frame.
```
etiennebacher, line 21:
```suggestion
* `describe_missing()`, to report on missing values in a data frame (#561).
```

R/describe_missing.R
```diff
@@ -0,0 +1,112 @@
+#' @title Describe Missing Values in Data According to Guidelines
+#'
+#' @description Provides a detailed description of missing values in a data frame.
+#' This function reports both absolute and percentage missing values of specified
```
etiennebacher, line 4:
```suggestion
#' This function reports both absolute number and percentage of missing values of specified
```

R/describe_missing.R
```diff
@@ -0,0 +1,112 @@
...
+#' variables.
+#'
+#' @inheritParams extract_column_names
+#' @param by Optional character string, indicating the names of one or more
+#' variables in the data frame. If supplied, the data will be split by these
+#' variables and summary statistics will be computed for each group. Useful
+#' for survey data by first reshaping the data to the long format.
```
etiennebacher, line 11: I think the last sentence is very specific and hard to understand. This usecase is part of the conversation so we have it in mind but it's a bit obscure for an external reader. It should be removed IMO.

R/describe_missing.R
```diff
@@ -0,0 +1,112 @@
...
+#' @param sort Logical. Whether to sort the result from highest to lowest
+#' percentage of missing data.
```
etiennebacher, line 13: Given this can be done with an extra `data_arrange()`, I don't think it's necessary to add this argument.

R/describe_missing.R
```diff
@@ -0,0 +1,112 @@
...
+#' @return A dataframe with the following columns:
+#'  - `variable`: Variables selected.
+#'  - `n_missing`: Number of missing values.
+#'  - `missing_percent`: Percentage of missing values.
+#'  - `complete_percent`: Percentage of non-missing values.
+#' @param ... Arguments passed down to other functions. Currently not used.
```
etiennebacher, line 19: Do we actually need `...`?

If we keep it, it should be positioned before `@return`

R/describe_missing.R
```diff
@@ -0,0 +1,112 @@
...
+#'   select = -1,
+#'   names_sep = "_",
+#'   names_to = c("dimension", "item")
+#' )
+#'
+#' describe_missing(
+#'   df_long,
+#'   select = -c(1, 3),
+#'   by = "dimension"
+#' )
```
etiennebacher, line 48: This fails with an unclear message if there are more than one variable in `by`. 

The way this argument works is also not very clear to me. For instance, I'd find it more natural if the `by` variables were used to return a list of dataframes, e.g.:

```
# group 1, subgroup 1
<output of describe_missing() for this particular (nested) group>

# group 1, subgroup 2
<output of describe_missing() for this particular (nested) group>

etc.
```

R/describe_missing.R
```diff
@@ -0,0 +1,112 @@
...
+#' variables in the data frame. If supplied, the data will be split by these
+#' variables and summary statistics will be computed for each group. Useful
+#' for survey data by first reshaping the data to the long format.
+#' @param sort Logical. Whether to sort the result from highest to lowest
+#' percentage of missing data.
+#' @return A dataframe with the following columns:
+#'  - `variable`: Variables selected.
+#'  - `n_missing`: Number of missing values.
+#'  - `missing_percent`: Percentage of missing values.
+#'  - `complete_percent`: Percentage of non-missing values.
```
etiennebacher, line 18: Should it also have the total number of obs for better comparison? Although this number would be repeated for all rows...

etiennebacher (changes requested):

R/describe_missing.R
bwiernik: What do you mean by nested structure here?

bwiernik: I think we should include `select` and `exclude` with their standard behavior and also include the current behavior as `composites`. This will let the function be usable for describing single-item measures as well as enable the advanced scale-level functionality

R/describe_missing.R
rempsyc: We don't need it, but I had added it because I pass it to `data_select()` for full compatibility, which has it as argument. That said, current documentation for `?data_select()` is 
> Arguments passed down to other functions. Mostly not used yet.

So I can either keep it or remove it, as you wish.

PR #563, opened 2024-11-21
Allow glue-styled pattern for `data_rename()`

strengejacke:
``` r
library(datawizard)
head(data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "formerly_{col}"))
#>                   formerly_mpg formerly_cyl formerly_disp
#> Mazda RX4                 21.0            6           160
#> Mazda RX4 Wag             21.0            6           160
#> Datsun 710                22.8            4           108
#> Hornet 4 Drive            21.4            6           258
#> Hornet Sportabout         18.7            8           360
#> Valiant                   18.1            6           225
head(data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "{col}_is_column_{n}"))
#>                   mpg_is_column_1 cyl_is_column_2 disp_is_column_3
#> Mazda RX4                    21.0               6              160
#> Mazda RX4 Wag                21.0               6              160
#> Datsun 710                   22.8               4              108
#> Hornet 4 Drive               21.4               6              258
#> Hornet Sportabout            18.7               8              360
#> Valiant                      18.1               6              225
```

<sup>Created on 2024-11-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/data_rename.R
+++ b/R/data_rename.R
@@ -24,10 +24,15 @@
 #'     in `pattern`. `pattern` and `replacement` must be of the same length.
 #'   - `NULL`, in which case columns are numbered in sequential order.
 #'   - A string (i.e. character vector of length 1) with a "glue" styled pattern.
-#'     Currently supported tokens are `{col}` (or `{name}`) and `{n}`. `{col}`
-#'     will be replaced by the column name, i.e. the corresponding value in
-#'     `pattern`. `{n}` will be replaced by the number of the variable that is
-#'     replaced. For instance,
+#'     Currently supported tokens are:
+#'     - `{col}` (or `{name}`), which will be replaced by the column name, i.e.
+#'       the corresponding value in `pattern`.
+#'     - `{n}` will be replaced by the number of the variable that is replaced.
+#'     - `{letter}` will be replaced by alphabetical letters in sequential order.
+#'     - Finally, the name of a user-defined object that is available in the
+#'       environment can be used. In this case,
+#'
+#'     An example for the use of tokens is...
 #'     ```r
 #'     data_rename(
 #'       mtcars,
@@ -35,8 +40,8 @@
 #'       replacement = "new_name_from_{col}"
 #'     )
 #'     ```
-#'     would returns new column names `new_name_from_am` and `new_name_from_vs`.
-#'     See 'Examples'.
+#'     ... which would return new column names `new_name_from_am` and
+#'     `new_name_from_vs`. See 'Examples'.
 #'
 #' If `pattern` is a named vector, `replacement` is ignored.
 #' @param rows Vector of row names.
@@ -66,6 +71,11 @@
 #' # Use glue-styled patterns
 #' head(data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "formerly_{col}"))
 #' head(data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "{col}_is_column_{n}"))
+#' head(data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "new_{letter}"))
+#'
+#' # User-defined glue-styled patterns from objects in environment
+#' x <- c("hi", "there", "!")
+#' head(data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "col_{x}"))
 #' @seealso
 #' - Functions to rename stuff: [data_rename()], [data_rename_rows()],
 #'   [data_addprefix()], [data_addsuffix()]
@@ -212,7 +222,7 @@ data_rename <- function(data,
     # prepare pattern
     column_name <- pattern[i]
     out[i] <- replacement
-    # replace first accepted token
+    # replace first pre-defined token
     out[i] <- gsub(
       "(.*)(\\{col\\})(.*)",
       replacement = paste0("\\1", column_name, "\\3"),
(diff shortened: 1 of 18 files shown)


strengejacke:
@DominiqueMakowski @easystats/core-team any suggestions for further patterns?

strengejacke:
the many rd-files are just due to lintr-updates, can be ignored when reviewing this PR.

strengejacke:
Rendering of help page:

## Positron

image

## RStudio

image

R/data_rename.R
```diff
@@ -10,18 +10,40 @@
...
+#'     vector using `<new name> = "<old name>"` and argument `replacement` will
+#'     be ignored then).
+#' @param replacement Character vector. Can be one of the following:
+#'   - A character vector that indicates the new names of the columns selected
+#'     in `pattern`. `pattern` and `replacement` must be of the same length.
+#'   - `NULL`, in which case columns are numbered in sequential order.
+#'   - A string (i.e. character vector of length 1) with a "glue" styled pattern.
+#'     Currently supported tokens are:
+#'     - `{col}` (or `{name}`), which will be replaced by the column name, i.e.
+#'       the corresponding value in `pattern`.
```
etiennebacher, line 29: I'd rather keep only one of them (but no preference between the two). This argument can already take various inputs, I don't think we should add aliases in possible values on top of that.

R/data_rename.R
```diff
@@ -10,18 +10,40 @@
...
+#'     - `{n}` will be replaced by the number of the variable that is replaced.
+#'     - `{letter}` will be replaced by alphabetical letters in sequential order.
```
etiennebacher, line 31: What happens if the user renames more than 26 columns?

R/data_rename.R
```diff
@@ -10,18 +10,40 @@
...
+#'     - Finally, the name of a user-defined object that is available in the
+#'       environment can be used.
```
etiennebacher, line 33: This should mention that this object cannot be named "col", "n", or "letter", otherwise it conflicts with the other tokens.

R/data_rename.R
```diff
@@ -122,14 +157,19 @@ data_rename <- function(data,
     }
   }
 
+  # check if we have "glue" styled replacement-string
+  glue_style <- length(replacement) == 1 &&
+    grepl("{", replacement, fixed = TRUE) &&
+    length(pattern) > 1
```
etiennebacher, line 163: I'm wondering if the `length(pattern) > 1` constraint is useful here. I think it may lead to subtle bugs if someone uses this function like this:
``` r
library(datawizard)
my_fun <- function(cols_to_rename) {
  data_rename(head(mtcars)[, 1:6], cols_to_rename, "new_{col}")
} 

# Token is used
names(my_fun(c("mpg", "drat")))
#> [1] "new_mpg"  "cyl"      "disp"     "hp"       "new_drat" "wt"

# Token is not used
names(my_fun("mpg"))
#> [1] "new_{col}" "cyl"       "disp"      "hp"        "drat"      "wt"
```

R/data_rename.R
```diff
@@ -167,6 +212,71 @@ data_rename <- function(data,
 }
 
 
+.glue_replacement <- function(pattern, replacement) {
+  # this function replaces "glue" tokens into their related
+  # real names/values. Currently, following tokens are accepted:
+  # - {col}/{name}: replacement is the name of the column (indicated in "pattern")
+  # - {n}: replacement is the number of the variable out of n, that should be renamed
```
etiennebacher, line 219: Can you add a comment for `{letter}` as well?

tests/testthat/test-data_rename.R
```diff
@@ -140,3 +140,31 @@ test_that("data_rename preserves attributes", {
...
+  new.env(),
+  test_that("data_rename glue-style, environment", {
+    data(mtcars)
+    x <- c("hi", "there", "!")
+    out <- data_rename(mtcars[1:3], c("mpg", "cyl", "disp"), "col_{x}")
+    expect_named(out, c("col_hi", "col_there", "col_!"))
+    expect_error(
+      data_rename(mtcars[1:3], c("mpg", "disp"), "col_{x}"),
+      regex = "The number of values"
+    )
```
etiennebacher, line 168: Can you add a check when `x` doesn't exist at all in the environment?

etiennebacher (changes requested):

R/data_rename.R
```diff
@@ -257,7 +266,14 @@ data_rename <- function(data,
...
         # evaluate token-object from the environment
-        values <- .dynEval(str2lang(gsub("\\{(.*)\\}", "\\1", token)))
+         values <- tryCatch(
+          .dynEval(str2lang(gsub("\\{(.*)\\}", "\\1", token))),
+          error = function(e) {
+            insight::format_error(paste0(
+              "The object `", token, "` was not found. Please check if it really exists."
+            ))
+          }
+        )
```
etiennebacher, line 276: `.dynEval()` has an argument `ifnotfound` that you can use instead of `tryCatch()`.

bwiernik:
Could we use `.col` or `.n` or `.letter` instead? I regularly see data frames with names "col" and "n" and am worried about potential collisions. Would also be consistent with tidyverse naming conventions where "special" column names/pronouns start with `.`

strengejacke:
I don't think it matters because it's only recognized when inside `{}`. You usually don't have column names like `{n}_times`, but rather `n_times`.

strengejacke:
I think it's even less problematic: the `{}` tokens are placeholders in the `rename` argument and don't refer to columns that should be selected, so we don't run into any problems here, or am I missing something?

bwiernik:
I wasn't sure how you had implemented the {} and whether it was evaluating the names inside {} (like glue does)

PR #565, opened 2024-11-27
Fix inconsitencies in docs

strengejacke:
Just fixing some minor inconsistencies in the docs.

### DIFF
--- a/R/extract_column_names.R
+++ b/R/extract_column_names.R
@@ -19,15 +19,16 @@
 #'   - a vector of negative integers, giving the positions counting from the
 #'     right (e.g., `-1` or `-1:-3`),
 #'   - one of the following select-helpers: `starts_with()`, `ends_with()`,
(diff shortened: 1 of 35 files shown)


strengejacke:
Since we inherit the docs for `select` in many functions, many rd-files are affected, but effectively you only need to review one rd-file.

R/extract_column_names.R
```diff
@@ -19,15 +19,16 @@
 #'   - a vector of negative integers, giving the positions counting from the
 #'     right (e.g., `-1` or `-1:-3`),
 #'   - one of the following select-helpers: `starts_with()`, `ends_with()`,
-#'     `contains()`, a range using `:` or `regex("")`. `starts_with()`,
+#'     `contains()`, a range using `:` or `regex()`. `starts_with()`,
```
etiennebacher, line 22:
```suggestion
#'     `contains()`, a range using `:`, or `regex()`. `starts_with()`,
```

PR #567, opened 2024-11-28
Check column names of dataframes?

### LINKED ISSUE #276: Check column names of dataframes?
Should this kind of check be generalized to all other functions for consistency? If so, we could make a helper `.check_dataframe_names()` (we could also suggest using `janitor::clean_names()`).

https://github.com/easystats/datawizard/blob/89e81930548aecf37e87968f83e5007dd3c37b8f/R/adjust.R#L78-L83

strengejacke:
Part of #276


R/adjust.R
```diff
@@ -75,12 +75,8 @@ adjust <- function(data,
...
                    regex = FALSE,
                    verbose = FALSE) {
-  if (!all(colnames(data) == make.names(colnames(data), unique = TRUE))) {
-    insight::format_warning(
-      "Bad column names (e.g., with spaces) have been detected which might create issues in many functions.",
-      "Please fix it (you can run `names(mydata) <- make.names(names(mydata))` for a quick fix)."
-    )
-  }
+  # make sure column names are syntactically valid
+  .check_dataframe_names(data, action = "error")
```
etiennebacher, line 79: Should we error here? This is not really wrong, it's "just" bad practice. I think this could break a bunch of code (could be done for 1.0 though).

I think we could also recommend `janitor::clean_names()`.

R/utils.R
```diff
@@ -47,6 +47,22 @@
 }
 
 
+#' Checks data framesfor syntactically valid column names
```
etiennebacher, line 50:
```suggestion
#' Checks dataframes for syntactically valid column names
```

R/adjust.R
strengejacke: `janitor::clean_names()` is included in the message. I think an error here is ok, because `adjust()` fails anyway with invalid column names, so why show a warning and an additional error?

etiennebacher: Right, my bad. I didn't know `adjust()` would fail with invalid column names so erroring is fine.

etiennebacher (approved):

etiennebacher:
I don't know if you want to expand this check to other functions in this PR or not, so feel free to merge.

strengejacke:
Let me think about which functions could get this check a bit... I revised the wording, because "Please fix it" is not always necessary, and maybe people want "invalid" column names. New wording is:
{E9EABE37-7AF0-454D-8F8F-72ADF3F7407D}

PR #568, opened 2024-11-28
Deprecate `pattern` in favour of `select` in `data_rename()`

### LINKED ISSUE #566: Use `select` instead of `pattern` in `data_rename()`?
I think when the function was initially written, it was very feature-limited. That's why we had the argument `pattern`, instead of `select`. Meanwhile, `data_rename()` is much more flexible, and the `pattern` argument came closer to what `select` typically does or can do.

Should we rename `pattern` into `select` then? Opinions?

@easystats/core-team @etiennebacher @DominiqueMakowski

etiennebacher:
Close #566

### DIFF
--- a/R/data_rename.R
+++ b/R/data_rename.R
@@ -10,7 +10,7 @@
 #'   pipe-workflow.
 #'
 #' @param data A data frame, or an object that can be coerced to a data frame.
-#' @param pattern Character vector.
+#' @param select Character vector.
 #'   - For `data_addprefix()` or `data_addsuffix()`, a character string, which
 #'     will be added as prefix or suffix to the column names.
 #'   - For `data_rename()`, indicates columns that should be selected for
@@ -51,6 +51,7 @@
 #' @param safe Do not throw error if for instance the variable to be
 #'   renamed/removed doesn't exist.
 #' @param verbose Toggle warnings and messages.
+#' @param pattern Deprecated. Use `select` instead.
 #' @param ... Other arguments passed to or from other functions.
 #'
 #' @return A modified data frame.
@@ -96,28 +97,34 @@
 #'
 #' @export
 data_rename <- function(data,
-                        pattern = NULL,
+                        select = NULL,
                         replacement = NULL,
                         safe = TRUE,
                         verbose = TRUE,
+                        pattern = NULL,
                         ...) {
+  if (!is.null(pattern)) {
+    .is_deprecated("pattern", "select")
+    select <- pattern
+  }
+
   # change all names if no pattern specified
-  if (is.null(pattern)) {
-    pattern <- names(data)
+  if (is.null(select)) {
+    select <- names(data)
   }
 
-  if (!is.character(pattern)) {
-    insight::format_error("Argument `pattern` must be of type character.")
+  if (!is.character(select)) {
+    insight::format_error("Argument `select` must be of type character.")
   }
 
-  # check if `pattern` has names, and if so, use as "replacement"
-  if (!is.null(names(pattern))) {
-    replacement <- names(pattern)
+  # check if `select` has names, and if so, use as "replacement"
+  if (!is.null(names(select))) {
+    replacement <- names(select)
   }
 
   # name columns 1, 2, 3 etc. if no replacement
   if (is.null(replacement)) {
-    replacement <- paste0(seq_along(pattern))
+    replacement <- paste0(seq_along(select))
   }
 
   # coerce to character
@@ -126,22 +133,22 @@ data_rename <- function(data,
   # check if `replacement` has no empty strings and no NA values
   invalid_replacement <- is.na(replacement) | !nzchar(replacement)
   if (any(invalid_replacement)) {
-    if (is.null(names(pattern))) {
-      # when user did not match `pattern` with `replacement`
+    if (is.null(names(select))) {
+      # when user did not match `select` with `replacement`
       msg <- c(
         "`replacement` is not allowed to have `NA` or empty strings.",
         sprintf(
-          "Following values in `pattern` have no match in `replacement`: %s",
-          toString(pattern[invalid_replacement])
+          "Following values in `select` have no match in `replacement`: %s",
+          toString(select[invalid_replacement])
         )
       )
     } else {
-      # when user did not name all elements of `pattern`
+      # when user did not name all elements of `select`
       msg <- c(
-        "Either name all elements of `pattern` or use `replacement`.",
+        "Either name all elements of `select` or use `replacement`.",
         sprintf(
-          "Following values in `pattern` were not named: %s",
-          toString(pattern[invalid_replacement])
+          "Following values in `select` were not named: %s",
+          toString(select[invalid_replacement])
         )
       )
     }
@@ -163,30 +170,30 @@ data_rename <- function(data,
   # check if we have "glue" styled replacement-string
   glue_style <- length(replacement) == 1 && grepl("{", replacement, fixed = TRUE)
 
-  if (length(replacement) > length(pattern) && verbose) {
+  if (length(replacement) > length(select) && verbose) {
     insight::format_alert(
       paste0(
-        "There are more names in `replacement` than in `pattern`. The last ",
-        length(replacement) - length(pattern), " names of `replacement` are not used."
+        "There are more names in `replacement` than in `select`. The last ",
+        length(replacement) - length(select), " names of `replacement` are not used."
       )
     )
-  } else if (length(replacement) < length(pattern) && verbose && !glue_style) {
+  } else if (length(replacement) < length(select) && verbose && !glue_style) {
     insight::format_alert(
       paste0(
-        "There are more names in `pattern` than in `replacement`. The last ",
-        length(pattern) - length(replacement), " names of `pattern` are not modified."
+        "There are more names in `select` than in `replacement`. The last ",
+        length(select) - length(replacement), " names of `select` are not modified."
       )
     )
   }
 
-  # if we have glue-styled replacement-string, create replacement pattern now
+  # if we have glue-styled replacement-string, create replacement select now
   if (glue_style) {
-    replacement <- .glue_replacement(pattern, replacement)
+    replacement <- .glue_replacement(select, replacement)
   }
 
-  for (i in seq_along(pattern)) {
+  for (i in seq_along(select)) {
     if (!is.na(replacement[i])) {
-      data <- .data_rename(data, pattern[i], replacement[i], safe, verbose)
+      data <- .data_rename(data, select[i], replacement[i], safe, verbose)
     }
   }
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 0.13.0.15
+Version: 0.13.0.16
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0003-1995-6531")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,12 @@
 # datawizard (development)
 
-BREAKING CHANGES
+DEPRECATIONS
 
-* Argument `drop_na` in `data_match()` is deprecated now. Please use `remove_na`
-  instead.
+* Argument `drop_na` in `data_match()` is deprecated now. Please use 
+  `remove_na` instead.
+
+* Argument `pattern` in `data_rename()` is deprecated. Please use `select` 
+  instead (#567).
 
 CHANGES
 
--- a/tests/testthat/test-attributes.R
+++ b/tests/testthat/test-attributes.R
@@ -136,7 +136,7 @@ test_that("convert_to_na, attributes preserved", {
 test_that("data_rename, attributes preserved", {
   x <- mtcars
   attr(x, "myattri") <- "I'm here"
-  x2 <- data_rename(x, pattern = "hp", replacement = "horsepower")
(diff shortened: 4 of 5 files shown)


R/text_format.R
```diff
@@ -6,7 +6,15 @@
 #' @param width Positive integer giving the target column width for wrapping
 #' lines in the output. Can be "auto", in which case it will select 90\% of the
 #' default width.
-#' @inheritParams data_rename
+#' @param pattern Character vector.
+#'   - For `data_addprefix()` or `data_addsuffix()`, a character string, which
+#'     will be added as prefix or suffix to the column names.
+#'   - For `data_rename()`, indicates columns that should be selected for
```
strengejacke, line 12: This argument no longer applies to `data_rename()`.

R/data_rename.R
```diff
@@ -10,23 +10,23 @@
 #'   pipe-workflow.
 #'
 #' @param data A data frame, or an object that can be coerced to a data frame.
-#' @param pattern Character vector.
+#' @param select Character vector.
 #'   - For `data_addprefix()` or `data_addsuffix()`, a character string, which
 #'     will be added as prefix or suffix to the column names.
```
strengejacke, line 15: For `data_addprefix()`, the argument is still named `pattern` (which is intended). We could just `@inheritParams extract_column_names` for the `select` docs, and then (like in `data_select()`) add a note like:

```
#' @details
#'
#' Specifically for `data_select()`, `select` can also be a named character
#' vector. In this case, the names are used to rename the columns in the
#' output data frame. See 'Examples'.
```

(replace `data_select()` with `data_rename()`)

R/data_rename.R
```diff
@@ -39,18 +39,19 @@
...
 #'     `new_name_from_vs`. See 'Examples'.
 #'
-#' If `pattern` is a named vector, `replacement` is ignored.
+#' If `select` is a named vector, `replacement` is ignored.
 #' @param rows Vector of row names.
 #' @param safe Do not throw error if for instance the variable to be
 #'   renamed/removed doesn't exist.
 #' @param verbose Toggle warnings and messages.
+#' @param pattern Deprecated. Use `select` instead.
 #' @param ... Other arguments passed to or from other functions.
```
strengejacke, line 55: Not deprecated for `data_addprefix()`.

R/data_rename.R
```diff
@@ -96,28 +97,34 @@
...
+    .is_deprecated("pattern", "select")
+    select <- pattern
+  }
+
   # change all names if no pattern specified
-  if (is.null(pattern)) {
-    pattern <- names(data)
+  if (is.null(select)) {
+    select <- names(data)
   }
```
strengejacke, line 114: I think at this point we can directly use the code from other functions, like `data_select()`, which calls `.select_nse()`. line 112 could be:

```r
  select <- .select_nse(
    select,
    data,
    exclude,
    ignore_case = ignore_case,
    regex = regex,
    allow_rename = TRUE,
    verbose = FALSE
  )
```

Note the `allow_rename = TRUE`, which still allows named vectors for renaming.

R/data_rename.R
```diff
@@ -96,28 +97,34 @@
...
 
-  if (!is.character(pattern)) {
-    insight::format_error("Argument `pattern` must be of type character.")
+  if (!is.character(select)) {
+    insight::format_error("Argument `select` must be of type character.")
   }
```
strengejacke, line 118: This could be removed, as `pattern` (or now: `select`) no longer needs to be a name as string.

strengejacke (changes requested):

R/data_rename.R
etiennebacher: Actually this would change several behavior / messages. For instance:
```r
data_rename(head(iris), select = c(test = "Species", "Sepal.Length"))
```
errored before because there is a missing replacement, but now `.select_nse()` returns a named vector with `Sepal.Length` as value and name, which means the rest of the function doesn't error anymore.

Also, the argument `safe` would be useless because `.select_nse()` returns an empty character vector if some column names are unknown, and therefore this argument is not reached anymore. Just want to discuss about those before going further with the changes.

strengejacke: Regarding `safe`: I'd say we can remove this argument, and set `verbose = TRUE` when calling `.select_nse()`. First, this will warn users about non-existent columns (`.warn_not_found()` is called) and second, we error later if number of variables to rename, and number of new names in `rename` do not match.

Regarding not all elements named: That's a good point. We could argue that we trust users that they know what they're doing. Or we could even extend into something like:

```r
data_rename(mtcars, select = c(test = "mpg", "gear", "hp), rename = "new_{n}")
```

which would use names (if not identical) as new value to rename, and for those where names = values (`gear` and `hp` here), we use the values specified in `rename` (though, if we allow to mix rename-values mixed with/without tokens, it's much more complicated - maybe we just say if users know what they do and let `data_rename(head(iris), select = c(test = "Species", "Sepal.Length"))` just happen...)

strengejacke: I think overall, the benefits of having the classic `select` functionality outweighs the new "limitations" we get.

strengejacke:
The docs are a bit messed up at the moment:
{8EA7D9FD-CD68-4199-9999-0FEAF59F2AC0}

`pattern` is not deprecated for `data_addprefix()` (and suffix), and the docs for `select` (for `data_addprefix()` is actually for the `pattern` argument. The `select` argument in `data_addprefix()` uses the normal select-helper functionality. We either have to split `data_addprefix()` and `data_addsuffix()` from `data_rename()` to have two different help-files, or, if we really use the classical `select` in `data_rename()`, we don't have to change much in the docs. Just `pattern` only refers to `data_addprefix()` and that's it.

etiennebacher:
I think it makes sense to split docs between `data_rename()` in one page and `data_addprefix()` and `data_addsuffix()` in another. We can use `@seealso` to redirect to the other.

R/data_rename.R
etiennebacher: > First, this will warn users about non-existent columns (.warn_not_found() is called) and second, we error later if number of variables to rename, and number of new names in rename do not match.

This was actually not the case. We only throw a message when `length(replacement) != length(select)`, which I think is not the right thing to do. I've changed it to an error, otherwise there's no replacement for `safe = TRUE` (the default).

> We could argue that we trust users that they know what they're doing. [...] `data_rename(head(iris), select = c(test = "Species", "Sepal.Length"))`

Actually I think using column indices as default column names if no replacement is provided is a weird behavior. Isn't a single number a "syntactically invalid name" as in #567? Do you have any usecase with this?

strengejacke: But why column indices? I think `select = c(test = "Species", "Sepal.Length")` returns a fully named vector, meaning that variables with no new names are "ignored":

``` r
datawizard::data_select(head(iris), select = c(test = "Species", "Sepal.Length"))
#>     test Sepal.Length
#> 1 setosa          5.1
#> 2 setosa          4.9
#> 3 setosa          4.7
#> 4 setosa          4.6
#> 5 setosa          5.0
#> 6 setosa          5.4
```

<sup>Created on 2024-11-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher: Right, I got confused with the behavior for unnamed vectors (e.g. just specifying `select`) where column indices are used:
``` r
datawizard::data_rename(head(iris), select = c("Species", "Sepal.Length"))
#>     2 Sepal.Width Petal.Length Petal.Width      1
#> 1 5.1         3.5          1.4         0.2 setosa
#> 2 4.9         3.0          1.4         0.2 setosa
#> 3 4.7         3.2          1.3         0.2 setosa
#> 4 4.6         3.1          1.5         0.2 setosa
#> 5 5.0         3.6          1.4         0.2 setosa
#> 6 5.4         3.9          1.7         0.4 setosa
```

So two things:
* I think we should keep the behavior in `main` and error if the vector is named but some elements don't have names
* using column indices when no replacement is provided seems like a bad practice and we could error because `length(select) > length(replacement)`

strengejacke: Yes, agreed. Only if `replacement` is `NULL` and no named vector in `select`, then sequential numbering is intended.

etiennebacher: Ah sorry I should have been clearer. For the second point, I meant that using column indices as replacement is a bad practice in all cases. For example, IMO this should error:
```r
datawizard::data_rename(head(iris), select = c("Species", "Sepal.Length"))
```
should error.

strengejacke: Yes, yes, I understood. But I think this was an intentional behavior, if I recall right:

Screenshot_20241130-154926

I think @DominiqueMakowski wanted that feature?

etiennebacher: > For a named vector, .select_nse() always returns a fully named vector. For non-named elements, the element value is used as name. So if we want to check whether some elements don't have names, we would do this separately, right?

This could be done before `.select_nse()`, like this:
```r
if (!is.null(names(select)) && any(names(select) == "")) {
    insight::format_error("When `select` is a named vector, all elements must be named.")
}
```
I think mixing behaviors, e.g. allowing this `data_rename(x, c("foo" = "Species", "Sepal.Length"), "new_{col}")` would lead to confusing code, which is why I'd rather detect early when some elements of `select` are named and others aren't.

> we would need to modify .warn_not_found() in a way it could either warn or error, if we want to error on not-found variables. 

That's possible to do, but it would complexify a bit more the internals of `.select_nse()`. Maybe we could just catch the warning in `data_rename()` and convert it to error? Or just remove `safe`? (although I can see when this is useful, even more since we have the glue-style replacement).

> I think @DominiqueMakowski wanted that feature?

I'd be curious to know in which context renaming columns "1", "2", etc. is useful.

strengejacke: > This could be done before .select_nse(), like this:

That's ok for me, but doesn't this evaluate the `select` argument? What if `select` is not a character vector?

> Or just remove safe?

I'd say we remove `safe` and warn/error on every unclear case.

> That's possible to do, but it would complexify a bit more the internals of .select_nse(). Maybe we could just catch the warning in data_rename() and convert it to error? 

I think it's less complex than catching a warning and throwing an error. We just need to pass an argument like `action` to `.warn_not_found()`, and there we have `format_alert(..., type = action)`. Much less code, or not?

> I'd be curious to know in which context renaming columns "1", "2", etc. is useful.

Not sure - @DominiqueMakowski ?

etiennebacher: > That's ok for me, but doesn't this evaluate the select argument? What if select is not a character vector?

Right, I've done another approach in [`bd0be81` (#568)](https://github.com/easystats/datawizard/pull/568/commits/bd0be816d67090d372c79f35bfb5234a5fa06b66)

> I think it's less complex than catching a warning and throwing an error. We just need to pass an argument like action to .warn_not_found(), and there we have format_alert(..., type = action). Much less code, or not?

It adds an arg in all internal functions of NSE, see [`f2a56af` (#568)](https://github.com/easystats/datawizard/pull/568/commits/f2a56af1b29b2aa967924b389b7557cc405be4cb)

I had removed the behavior of using column indices for names, but I'll fix the remaining tests and examples once I know if we keep this behavior or not.

strengejacke: > It adds an arg in all internal functions of NSE

Right, it's indeed in more places than I thought...

> Right, I've done another approach in bd0be81 (#568)

Looks good, but could it be shorter? Instead of the for-loop, compare all names against elements:

```r
if (!is.null(names(select)) && any(names(select) == select)) {
...
}
```

strengejacke:
Changes look good to me, I think it's just that tests need to be updated?

R/data_rename.R
etiennebacher: Done in [`62631e6` (#568)](https://github.com/easystats/datawizard/pull/568/commits/62631e6ec9037fdae32efe58c5b4975ca76add68)

strengejacke:
> Yes I'm just waiting to see if we keep using column indices as valid replacements or not.

I'd say we can remove this. It's easy to do `colnames(x) <- seq_len(ncol(x))`. We also don't use this type of replacement in any easystats-packages, if I haven't missed something.

strengejacke:
@etiennebacher I think I updated all packages in _easystats_ that are affected by the new `data_rename()` changes. One thing I thought of was if we could/should keep the `safe` argument? Based on its value, we could set `ifnotfound` either to `"warn"` or `"error"`, which should restore the previous behaviour. WDYT?

etiennebacher:
Right, now that we have `ifnotfound` we can do that, I'll open a PR

etiennebacher:
Actually it's not that easy because if `.select_nse()` warns only, it doesn't include the unknown column name in the output. So for instance, if we do `data_rename(iris, "FakeCol", "length")` then `.select_nse()` returns `character(0)` and therefore we still error with `There are more names in 'replacement' than in
  'select'.`

strengejacke:
Hm, true. Or, if `safe = TRUE`, we just skip this check as well? Meaning that `safe` will "intersect" all valid columns and their replacement values, and use that for rename. Any values in `replacement` without a match in `select` will be ignored. For `safe = FALSE`, we set `ifnotfound = "error"`, *and* include the check for unequal lengths.

etiennebacher:
>  Or, if safe = TRUE, we just skip this check as well? Meaning that safe will "intersect" all valid columns and their replacement values, and use that for rename. 

But we don't know the position of the wrong column names since it's not returned by `.select_nse()`. For instance if we do this:
```r
data_rename(iris, c("Sepal.Length", "foo"), c("rep1", "rep2"))
```
then `.select_nse()` returns `"Sepal.Length"`, but how do we know which replacement we should use? We can't check the original `select` since it could be `regex()` for instance. I don't see a good way to enable this. Is the `safe` argument really important?

strengejacke:
Ok, agree. Then we don't need to change anything and deprecate the `safe` argument, as we now do.

PR #573, opened 2024-12-14
New argument `zeros` in `ranktransform()`

### LINKED ISSUE #570: Improving `ranktransform(sign = TRUE)`
Currently, if `x` contains 0s, these are marked as NAs.

However, there might be situations where this is unwanted (see https://github.com/easystats/effectsize/issues/502).

We can add a `zeros=` argument, that can take the values:
- `"na"` - the current method.
- `"signrank"` - keep them during the ranking and mark them as 0.

```R
if (zeros == "na") {
  out <- rep(NA, length(x))
  ZEROES <- x == 0
  out[!ZEROES] <- sign(x[!ZEROES]) * rank(abs(x[!ZEROES]), 
                                          ties.method = method,
                                          na.last = "keep")
} else if (zeros == "signrank") {
  out <- sign(x) * rank(abs(x), ties.method = method, na.last = "keep")
}
```

etiennebacher:
Close #570. @mattansb can you double check this? 

I also noticed there are no tests for the `method` argument. I think this came from `effectsize` so if you already have a suite of tests for this, we could add it here.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 0.13.0.18
+Version: 0.13.0.19
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0003-1995-6531")),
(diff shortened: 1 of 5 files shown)


mattansb:
LGTM!

I will use this in https://github.com/easystats/effectsize/issues/502 in the next cycle (:

mattansb (approved):

PR #579, opened 2024-12-30
Append results of `demean()` to original df?

### LINKED ISSUE #578: Append results of `demean()` to original df?
99% of the time was want to append the results of the `demean()` to the original data frame. Can we have that as an option (or maybe even the default)?

strengejacke:
Fixes #578

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,12 @@ BREAKING CHANGES AND DEPRECATIONS
   multiple tables is returned. Furthermore, `print_html()` did not work, which
   was also fixed now.
 
+* `demean()` (and `degroup()`) gets an `append` argument that defaults to `TRUE`,
+  to append the centered variabled to the original data frame, instead of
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -21,6 +21,12 @@ BREAKING CHANGES AND DEPRECATIONS
   multiple tables is returned. Furthermore, `print_html()` did not work, which
   was also fixed now.
 
+* `demean()` (and `degroup()`) gets an `append` argument that defaults to `TRUE`,
```
mattansb, line 24: This will require also updating the examples across easystats:

- the `demean` vignette in `{parameters}`
- `parameters:::.std_info_pseudo()`
- tests in `{parameters}`

strengejacke: Right, but the datawizard update will break parameters anyway

mattansb (approved):

Issue #576, opened 2024-12-18
data_rename() should accept lists


DominiqueMakowski:
Would be a useful feature

```r
datawizard::data_rename(mtcars, list(mpg="MPG", cyl="CYL"))
```

etiennebacher:
There is already support for named character vectors of the form `new_name = "old_name"`:
```r
data_rename(mtcars, c(MPG="mpg", CYL="cyl"))
```
Is adding named list support useful?

strengejacke:
I think a named vector is sufficient.

DominiqueMakowski:
If the support is low-clost (basically simply `unlist()` if patterns is list) I'd add it 🤷

strengejacke:
The new `select` argument in `data_rename()` is processed by the internal `.select_nse()` machinery, which seems to support lists, but not named lists. Not sure how hard it would be to add support for named lists?

etiennebacher:
I took a look and it would introduce more complexity in `.select_nse()`. The problem stems from 2 things:

* in `data_rename()`, we allow named vectors of the form `c(new_name = "old_name")`. Therefore for consistency we'd need to support `list(new_name = "old_name")` (and not `list(old_name = "new_name")` as in the original post.
* currently we handle named lists where the name of the element is the current variable name, e.g. `cyl = 1` (this is used in `convert_na_to()` for instance). Therefore, depending on which function calls `.select_nse()`, we'd need to use the names of the list or the values of the list as old variable names. `.select_nse()` is already hard to understand and maintain so I'd rather not add this.

A workaround for your usecase is:

``` r
library(datawizard)

to_rename <- list("MPG" = "mpg", "CYL" = "cyl")

data_rename(head(mtcars), unlist(to_rename))
#>                    MPG CYL disp  hp drat    wt  qsec vs am gear carb
#> Mazda RX4         21.0   6  160 110 3.90 2.620 16.46  0  1    4    4
#> Mazda RX4 Wag     21.0   6  160 110 3.90 2.875 17.02  0  1    4    4
#> Datsun 710        22.8   4  108  93 3.85 2.320 18.61  1  1    4    1
#> Hornet 4 Drive    21.4   6  258 110 3.08 3.215 19.44  1  0    3    1
#> Hornet Sportabout 18.7   8  360 175 3.15 3.440 17.02  0  0    3    2
#> Valiant           18.1   6  225 105 2.76 3.460 20.22  1  0    3    1
```

strengejacke:
[...]

PR #580, opened 2025-01-03
`data_rename`: mention use of `unlist()` for named lists

etiennebacher:
https://github.com/easystats/datawizard/issues/576#issuecomment-2569583348


strengejacke:
Ok, let's see if the lintr commit breaks anything, if not, I'd say we can merge this PR (and if it does, revert the last commit and merge the PR).

PR #586, opened 2025-02-14
Fix `data_arrange()` for 1-column data frames

strengejacke:


etiennebacher (approved):

etiennebacher:
That's a weird failure, `Hmisc` hasn't been updated since early January and I think its R 4.2 requirement is older than that.

Issue #587, opened 2025-02-26
Errors in CRAN checks


etiennebacher:
To fix before 2025-03-12: https://cran.r-project.org/web/checks/check_results_datawizard.html

etiennebacher:
Likely due to https://github.com/wch/r-source/commit/eb6b32bd6502b96e4bae4ec6443474d3f1b97762

PR #588, opened 2025-02-26
Fix R-devel error due to change in `grep()`

etiennebacher:
Fix for #587


strengejacke:
I edited the initial post, so this PR is recognized in the related issue ("Fix for..." -> "Fixes...")

etiennebacher:
Thanks but that was on purpose, I want to close the issue when I make the new release (not necessarily today)

PR #590, opened 2025-03-05
Test of modelbased devel verison

strengejacke:
_modelbased_ is about to be submitted, _insight_ and _parameters_ were just updated on CRAN, so we could check whether tests still pass.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -76,3 +76,4 @@ RoxygenNote: 7.3.2
 Config/testthat/edition: 3
 Config/testthat/parallel: true
 Config/Needs/website: easystats/easystatstemplate
+Remotes: easystats/modelbased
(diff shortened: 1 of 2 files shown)


strengejacke:
@etiennebacher Tests pass for the latest _parameters_ and for forthcoming _modelbased_ (current devel, to be submitted soon). Since _datawizard_ has a closer deadline than _modelbased_, we should probably set minimum requirements for failing tests (checks are currently running while I'm writing this), increasing modelbased minimum to 0.10.0, and parameters to 0.24.2 for failing tests.

strengejacke:
Ok, seems like no further action needed. I thought I saw something on the CRAN website, but maybe I was actually on the modelbased website, not datawizard 😄

PR #598, opened 2025-04-01
Only print `Variable` column if it exists

strengejacke:


etiennebacher (changes requested):
Thanks, can we have a test for this?

PR #601, opened 2025-04-04
Give informative error in `data_codebook()` when `select` matched no variables.

strengejacke:
Was:

``` r
data(iris)
datawizard::data_codebook(iris, select = starts_with("abc"))
#> Error in names(out) <- replace(names(out), names(out) == pattern[i], replacement[i]): attempt to set an attribute on NULL
```

Is:

``` r
data(iris)
datawizard::data_codebook(iris, select = starts_with("abc"))
#> NULL
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # datawizard (devel)
 
+CHANGES
+
+* `data_codebook()` gives an informative warning when no column names matched
+  the selection pattern.
+
 BUG FIXES
 
 * Fixed bug in `data_to_wide()`, where new column names in `names_from` were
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -1,5 +1,10 @@
 # datawizard (devel)
 
+CHANGES
+
+* `data_codebook()` gives an informative warning when no column names matched
+  the selection pattern.
```
```suggestion
* `data_codebook()` gives an informative warning when no column names matched
  the selection pattern (#601).
```

etiennebacher (approved):
Thanks, LGTM, can you just fix the lintr workflow?

The other CI failure should be addressed in a PR of its own.

PR #603, opened 2025-04-08
Remove deprecated arguments

strengejacke:


etiennebacher:
I'd have to check in more detail but 4 months is not that old for an argument deprecation. Also this would be a breaking change so it would be better to use a minor version bump rather than a patch version for this.

strengejacke:
> Also this would be a breaking change so it would be better to use a minor version bump rather than a patch version for this.

Yes, but we can set the final version number just before release. For the devel versions, it's important to have at least an increase for the fourth place, to make sure `easystats::install_latest()` works.

strengejacke:
> but 4 months is not that old for an argument deprecation

Yes, usually not. But we deprecated this with version 1, so we already had that "API" change, which would justify the clean-up ;-) But I leave it to you to keep the deprecation messages longer and remove the arguments later.

etiennebacher:
I'd like to have at least 6 months of deprecation, i.e. wait until early July, before removing them, so putting this on draft and this won't be a blocker for a potential release next week (cf #615).

strengejacke:
@gemini-code-assist summary

strengejacke:
Probably doesn't work on demand with Gemini code assist?

PR #605, opened 2025-04-10
New handling of expressions in `data_modify()`, introducing `as_expr()` helper function

strengejacke:
```r
library(datawizard)

data_modify(iris, as_expression("sepwid = 2 * Sepal.Width")) |> head()

data_modify(
    iris,
    as_expression(c("sepwid = 2 * Sepal.Width", "seplen = 5 * Sepal.Length"))
) |> head()

data_modify(iris, sepwid = as_expression("2 * Sepal.Width")) |> head()

data_modify(
    iris,
    sepwid = as_expression("2 * Sepal.Width"),
    seplen = {"5 * Sepal.Length"}
) |> head()

e <- "sepwid = 2 * Sepal.Width"
data_modify(iris, as_expression(e)) |> head()

e <- c("sepwid = 2 * Sepal.Width", "seplen = 5 * Sepal.Length")
data_modify(iris, as_expression(e)) |> head()

e <- "2 * Sepal.Width"
data_modify(iris, sepwid = as_expression(e)) |> head()

e <- "2 * Sepal.Width"
data_modify(iris, sepwid = as_expression(e), seplen = 5 * Sepal.Length) |> head()

e <- "2 * Sepal.Width"
f <- "half_petal = 0.5 * Petal.Length"
a <- "string"
num <- 1:5
data_modify(
    iris,
    sepwid = as_expression(e),
    seplen = 5 * Sepal.Length,
    {f},
    new_var = a,
    new_num = num,
    new_var2 = "ho",
    new_num2 = 4:6,
    Sepal.Length = NULL,
    Petal.Length = NULL,
    Sepal.Width = NULL,
    Petal.Width = NULL
) |> head()


d <- data.frame()
for (param in letters[c(1, 2, 5)]) {
  out <- data.frame(x = as.numeric(as.factor(param)))
  out <- data_modify(out, Parameter = param)
  d <- rbind(out, d)
}
d

data_modify(iris, sepwid = "2 * Sepal.Widht") |> head()

# errors

e <- "sepwid = 2 * Sepal.Widht"
data_modify(iris, as_expression(e))

data_modify(iris, as_expression("sepwid = 2 * Sepal.Widht"))

data_modify(iris, sepwid = 2 * Sepal.Widht)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,8 +17,8 @@ BREAKING CHANGES
 * `data_modify()` now always uses values of a vector for a modified or newly
   created variable, and no longer tries to detect whether a character value
   possibly contains an expression. To allow expression provided as string (or
-  character vectors), use the helper-function `as_expression()`. Only literal
-  expressions or strings wrapped in `as_expression()` will be evaluated as
+  character vectors), use the helper-function `as_expr()`. Only literal
+  expressions or strings wrapped in `as_expr()` will be evaluated as
   expressions, everything else will be treated as vector with values for new
   variables (#605)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,10 @@ BREAKING CHANGES
   requested has changed. Now, for each centrality measure a confidence interval
   is calculated (#617).
 
+* `data_modify()` now treats values of variables used as expression on the RHS
+  as value for a newly created variable, and no longer copies data from a
+  variable if that value referred to the name of an existing variable.
+
 CHANGES
 
 * `display()` is now re-exported from package *insight*.
@@ -27,10 +31,6 @@ CHANGES
 
 * `data_to_long()` now errors when columns selected to reshape do not exist in
   the data, to avoid nonsensical results that could be missed (#602).
-
-* `data_modify()` now treats values of variables used as expression on the RHS
-  as value for a newly created variable, if that value does not refer to the
-  name of an existing variable.
 
 * New argument `by` in `describe_distribution()` (#604).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -38,8 +38,6 @@ CHANGES
 
 * New argument `by` in `describe_distribution()` (#604).
 
-* New argument `by` in `data_modify()` (#605).
-
 * `describe_distribution()` now gives informative errors when column names
   in the input data frame conflict with column from the output table (#612).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,9 +14,13 @@ BREAKING CHANGES
   requested has changed. Now, for each centrality measure a confidence interval
   is calculated (#617).
 
-* `data_modify()` now treats values of variables used as expression on the RHS
-  as value for a newly created variable, and no longer copies data from a
-  variable if that value referred to the name of an existing variable.
+* `data_modify()` now always uses values of a vector for a modified or newly
+  created variable, and no longer tries to detect whether a character value
+  possibly contains an expression. To allow expression provided as string (or
+  character vectors), use the helper-function `as_expression()`. Only literal
+  expressions or strings wrapped in `as_expression()` will be evaluated as
+  expressions, everything else will be treated as vector with values for new
+  variables (#605)
 
 CHANGES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@ BREAKING CHANGES
   character vectors), use the helper-function `as_expr()`. Only literal
   expressions or strings wrapped in `as_expr()` will be evaluated as
   expressions, everything else will be treated as vector with values for new
-  variables (#605)
+  variables (#605).
 
 CHANGES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -38,6 +38,8 @@ CHANGES
 
 * New argument `by` in `describe_distribution()` (#604).
 
+* New argument `by` in `data_modify()` (#605).
+
 * `describe_distribution()` now gives informative errors when column names
   in the input data frame conflict with column from the output table (#612).
 
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -185,11 +185,13 @@ data_modify.data.frame <- function(data, ..., .if = NULL, .at = NULL, .modify = 
           symbol_string <- insight::trim_ws(unlist(strsplit(symbol_string, ",", fixed = TRUE), use.names = FALSE))
         }
         # check if we have any symbols instead of strings as expression
-        for (s in seq_along(symbol_string)) {
-          if (!grepl("\"", symbol_string[s], fixed = TRUE)) {
-            symbol_string[s] <- .dynEval(str2lang(symbol_string[s]), data = data)
+        symbol_string <- unlist(lapply(symbol_string, function(s) {
+          if (!grepl("\"", s, fixed = TRUE)) {
+            .dynEval(str2lang(s), data = data)
+          } else {
+            s
           }
-        }
+        }), use.names = FALSE)
         # remove quotes from strings
         symbol_string <- gsub("\"", "", symbol_string)
         # check whether we have exact one = sign.
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -28,11 +28,16 @@
 #'   a <- "Sepal.Width = center(Sepal.Width)"
 #'   data_modify(iris, as_expr(a))
 #'   ```
+#'   Note that `as_expr()` is no real function, which cannot be used outside
+#'   of `data_modify()`, and hence it is not exported nor documented. Rather,
+#'   it is only used for internally processing expressions.
 #' - Using `NULL` as right-hand side removes a variable from the data frame.
 #'   Example: `Petal.Width = NULL`.
 #' - For data frames (including grouped ones), the function `n()` can be used to
 #'   count the number of observations and thereby, for instance, create index
-#'   values by using `id = 1:n()` or `id = 3:(n()+2)` and similar.
+#'   values by using `id = 1:n()` or `id = 3:(n()+2)` and similar. Note that,
+#'   like `as_expr()`, `n()` is also no true function and cannot be used outside
+#'   of `data_modify()`.
 #'
 #' Note that newly created variables can be used in subsequent expressions,
 #' including `.at` or `.if`. See also 'Examples'.
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -343,21 +343,7 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     #
     # in this case, we need to evaluate the symbol (i.e. convert symbol string
     # into a language expression and then evaluate)
-    symbol_string <- unlist(lapply(symbol_string, function(symbol_element) {
-      if (!startsWith(symbol_element, "\"")) {
-        return_value <- .dynEval(str2lang(symbol_element))
-        # dynEval might fail if we don't look in data - sanity check
-        if (identical(return_value, symbol_element)) {
-          return_value <- .dynEval(str2lang(symbol_element), data = data)
-        }
-        return_value
-      } else {
-        symbol_element
-      }
-    }), use.names = FALSE)
-    # now we should have the expression as character string. Next, we
-    # # remove quotes from strings
-    symbol_string <- gsub("^\"(.*)\"$", "\\1", symbol_string)
+    symbol_string <- .evaluate_expression_in_string(symbol_string, data)
     # check whether we have exact one = sign. We need to have a name definition,
     # i.e. something like "var = a+b" - if the string has no "=" sign, name is
     # definitely missing
@@ -426,22 +412,10 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     }
     # here we found an expression token - convert string into a regular expression
     if (!is.null(symbol_string)) {
-      # check if we have any symbols instead of strings as expression - evaluate
-      # and convert result into expression
-      symbol_string <- unlist(lapply(symbol_string, function(symbol_element) {
-        if (!startsWith(symbol_element, "\"")) {
-          return_value <- .dynEval(str2lang(symbol_element))
-          # dynEval might fail if we don't look in data - sanity check
-          if (identical(return_value, symbol_element)) {
-            return_value <- .dynEval(str2lang(symbol_element), data = data)
-          }
-          return_value
-        } else {
-          symbol_element
-        }
-      }), use.names = FALSE)
+      # check if we have any symbols instead of strings as expression
+      symbol_string <- .evaluate_expression_in_string(symbol_string, data)
       # remove quotes from strings and save symbol name
-      symbol_string <- .fix_quotes(gsub("^\"(.*)\"$", "\\1", symbol_string))
+      symbol_string <- .fix_quotes(symbol_string)
       symbol_name <- names(dots)[i]
       # convert string into language and replace in dots
       return_value <- try(str2lang(symbol_string), silent = TRUE)
@@ -464,6 +438,31 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
 
 
 # helper -------------
+
+
+.evaluate_expression_in_string <- function(symbol_string, data) {
+  # check if we have any symbols instead of strings as expression, e.g.
+  # xpr <- "sepwid = 2 * Sepal.Width"
+  # data_modify(iris, as_expr(xpr))
+  #
+  # in this case, we need to evaluate the symbol (i.e. convert symbol string
+  # into a language expression and then evaluate)
+  symbol_string <- unlist(lapply(symbol_string, function(symbol_element) {
+    if (!startsWith(symbol_element, "\"")) {
+      return_value <- .dynEval(str2lang(symbol_element))
+      # dynEval might fail if we don't look in data - sanity check
+      if (identical(return_value, symbol_element)) {
+        return_value <- .dynEval(str2lang(symbol_element), data = data)
+      }
+      return_value
+    } else {
+      symbol_element
+    }
+  }), use.names = FALSE)
+  # now we should have the expression as character string. Next, we
+  # # remove quotes from strings
+  gsub("^\"(.*)\"$", "\\1", symbol_string)
+}
 
 
 .fix_quotes <- function(symbol_string) {
--- a/R/data_modify.R
+++ b/R/data_modify.R
@@ -17,7 +17,7 @@
 #'   a <- "abc"
 #'   data_modify(iris, var_abc = a) # var_abc contains "abc"
 #'   ```
-#' - An expression can also be provided as string and wrapped in 
+#' - An expression can also be provided as string and wrapped in
 #' `as_expression()`. Example:
 #'   ```r
 #'   data_modify(iris, as_expression("Sepal.Width = center(Sepal.Width)"))
@@ -182,7 +182,7 @@ data_modify.data.frame <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     # a <- "sepwid = 2 * Sepal.Width"
     # data_modify(iris, as_expression(a))
     #
-    dots <- .extract_unnamed_expressions(dots, data)
+    dots <- .process_unnamed_expressions(dots, data)
 
     # next, we check for named expression-tags and convert these into regular
     # expressions, e.g.
@@ -191,7 +191,7 @@ data_modify.data.frame <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     # a <- "2 * Sepal.Width"
     # data_modify(iris, sepwid = as_expression(a))
     #
-    dots <- .extract_named_expressions(dots, data)
+    dots <- .process_named_expressions(dots, data)
 
     for (i in seq_along(dots)) {
       # create new variable
@@ -241,7 +241,7 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     # a <- "sepwid = 2 * Sepal.Width"
     # data_modify(iris, as_expression(a))
     #
-    dots <- .extract_unnamed_expressions(dots, data)
+    dots <- .process_unnamed_expressions(dots, data)
 
     # next, we check for named expression-tags and convert these into regular
     # expressions, e.g.
@@ -250,7 +250,7 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
     # a <- "2 * Sepal.Width"
     # data_modify(iris, sepwid = as_expression(a))
     #
-    dots <- .extract_named_expressions(dots, data)
+    dots <- .process_named_expressions(dots, data)
   }
 
   # create new variables as dummys, do for-loop works
@@ -280,7 +280,7 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify = 
 
 # expression processing ----------------------------------------------------
 
-.extract_unnamed_expressions <- function(dots, data) {
+.process_unnamed_expressions <- function(dots, data) {
   # dots are only unnamed, when the full expression is saved in a string,
   # e.g. data_modify(iris, as_expression("sepwid = 2 * Sepal.Width")).
   # Thus, we know we *have to* find an expression here, and the string value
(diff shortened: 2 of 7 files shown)


strengejacke:
Sure. Briefly, you could pass a variable as expression in `data_modify()` that can be evaluated, e.g. it contains a string (with a name of another variable, or an expression as string...) like

```r
a <- "Sepal.Width"
data_modify(iris, new_sepal = a) # new_sepal = Sepal.Width
```

@DominiqueMakowski had a situation (for-loop), where he wanted to create a new variable, and the variable used as expression was a string, but that string was not referring to a variable name nor did it contain an expression, but was the value that should be assigned to the newly created variable (see example above). This was not possible, because `data_modify()` considered the value as name of an object.

This PR now checks:
- can a string be interpreted as expression?
- if not, because variable not found, is the string actually a "value"?
- if yes, use it. else, we have an "object not found" issue.

R/data_modify.R
```diff
@@ -10,11 +10,24 @@
...
 #'   to the name of the new variable, while the right-hand side represent the
 #'   values of the new variable. Example: `Sepal.Width = center(Sepal.Width)`.
-#' - A sequence of string values, representing expressions.
-#' - A variable that contains a string representation of the expression. Example:
+#' - A sequence of string values, representing expressions (see 'Examples').
+#' - A variable that contains a string of a variable name. The values from this
+#'   variable will be copied. Example:
+#'   ```r
+#'   a <- "Sepal.Width"
+#'   data_modify(iris, new_sepal = a) # new_sepal = Sepal.Width
```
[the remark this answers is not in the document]
strengejacke: Hm, I see the point. Maybe it's better to keep just one behavior, but which?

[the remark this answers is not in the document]
strengejacke: @DominiqueMakowski any thoughts?

strengejacke:
I think this PR not necessarily needs to be merged prior to a new release.

I personally tend to change the current behaviour, so that values in variables are used for a newly created variable (new behaviour), instead of searching for column names (old behaviour), so we would allow:

```r
a <- "abc"
data_modify(iris, var_abc = a) # var_abc contains "abc"
```

However, since we allow expressions as strings, like `c("SW_double = 2 * Sepal.Width", "SW_fraction = SW_double / 10")`, we have to distinguish these from "values". We do this in:

```r
new_variable <- try(with(data, eval(symbol)), silent = TRUE)
```

But this would _also_ evaluate `symbol` if it is a column name (which we no longer want). Thus, we must:

1. check if `symbol` is a valid column name in the data. If yes, we want it as new value, we no longer want to copy the variable indicated in `symbol` (i.e. we treat the variable name as if it was the new value for the new variable)
2. if `symbol` is no column name, we can then evaluate with `try(with(data, eval(symbol)), silent = TRUE)`, checking if the string contains an expression.
3. if not, we are sure we have a value, and just use this.

strengejacke:
@etiennebacher @DominiqueMakowski ok, I'm done with revising this PR.

The main purpose: if a variables contains a character value, this value can be used for the newly created variable.

Example:

```r
a <- "abc"
data_modify(iris, var_abc = a) # var_abc contains "abc"
```

If by coincidence the value also refers to a column name in the data, it is still treated as value, the column from the data is no longer copied (breaking change).

Example:

``` r
library(datawizard)
a <- "Sepal.Width"
data_modify(iris, double_SepWidth = a) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species double_SepWidth
#> 1          5.1         3.5          1.4         0.2  setosa     Sepal.Width
#> 2          4.9         3.0          1.4         0.2  setosa     Sepal.Width
#> 3          4.7         3.2          1.3         0.2  setosa     Sepal.Width
#> 4          4.6         3.1          1.5         0.2  setosa     Sepal.Width
#> 5          5.0         3.6          1.4         0.2  setosa     Sepal.Width
#> 6          5.4         3.9          1.7         0.4  setosa     Sepal.Width
```

Problem: we also allow expressions to be passed as character string.

Example:

``` r
library(datawizard)
a <- "2 * Sepal.Width"
data_modify(iris, double_SepWidth = a) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species double_SepWidth
#> 1          5.1         3.5          1.4         0.2  setosa             7.0
#> 2          4.9         3.0          1.4         0.2  setosa             6.0
#> 3          4.7         3.2          1.3         0.2  setosa             6.4
#> 4          4.6         3.1          1.5         0.2  setosa             6.2
#> 5          5.0         3.6          1.4         0.2  setosa             7.2
#> 6          5.4         3.9          1.7         0.4  setosa             7.8

data_modify(iris, double_SepWidth = 2 * Sepal.Width) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species double_SepWidth
#> 1          5.1         3.5          1.4         0.2  setosa             7.0
#> 2          4.9         3.0          1.4         0.2  setosa             6.0
#> 3          4.7         3.2          1.3         0.2  setosa             6.4
#> 4          4.6         3.1          1.5         0.2  setosa             6.2
#> 5          5.0         3.6          1.4         0.2  setosa             7.2
#> 6          5.4         3.9          1.7         0.4  setosa             7.8
```

The problem is how to distinguish a _value_ from an _expression_, when passed as character? We could say: if it can be evaluated, it is an expression, else, a value. But: what if the expression contains a typo?

``` r
library(datawizard)
a <- "2 * Sepal.Widht"
data_modify(iris, double_SepWidth = a) |> head()
#> Error: There was an error in the first expression. Variable "Sepal.Widht" was
#>   not found in the dataset or in the environment.
#>   Did you mean one of "Sepal.Length", "Sepal.Width" or "Petal.Width"?
```

This works, because if the evaluation of the expression `"2 * Sepal.Widht"` fails, `.is_valid_value()` checks for certain characters in that string that _might_ indicate that we have an expression, and no value. Whenever following characters are present: `<>+-=/*()` it is considered as expression, so there might be a typo if it fails to evaluate. If none of these characters are in the expression, it is considered as value.

``` r
library(datawizard)
a <- "new1"
data_modify(iris, double_SepWidth = a) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species double_SepWidth
#> 1          5.1         3.5          1.4         0.2  setosa            new1
#> 2          4.9         3.0          1.4         0.2  setosa            new1
#> 3          4.7         3.2          1.3         0.2  setosa            new1
#> 4          4.6         3.1          1.5         0.2  setosa            new1
#> 5          5.0         3.6          1.4         0.2  setosa            new1
#> 6          5.4         3.9          1.7         0.4  setosa            new1

a <- "new-1"
data_modify(iris, double_SepWidth = a)
#> Error: There was an error in the first expression. Non-numeric argument to
#>   binary operator. Possibly misspelled or not yet defined?
```

This is not perfect, but currently the best solution I have to allow both _values_ and _expressions_ as character vectors, which is a convenient way to avoid "undefined global variable" warnings in R packages when using `data_modify()`...

strengejacke:
Would we add a "real function", or just parse the expression into a string and check for `"^eval_as_expr\\((.*)\\)$"` or something like that?

etiennebacher:
I think it would be a real function that doesn't work when called outside of `data_modify()` (and possibly `data_filter()`?), similar to as `starts_with()` doesn't work outside a subset of `dplyr` functions.

strengejacke:
But for `select`, we don't have those functions, we just evaluate the expression:

```r
.regex_select_helper <- function() {
  "(starts\\_with|ends\\_with|col\\_ends\\_with|contains|regex)"
}
```

I thought we could do this here as well?

strengejacke:
> Since we're having a breaking change anyway, I'd be more comfortable with having a dedicated function to a evaluate string as an expression in the data context (although I always have [this SO post](https://stackoverflow.com/a/40164111/11598948) in my head when talking about evaluating strings), so that we have for instance:
> 
> ```r
> a <- "2 * Sepal.Width"
> data_modify(iris, double_SepWidth = a) # gives a column full of "2 * Sepal.Width"
> 
> data_modify(iris, double_SepWidth = eval_as_expr(a)) # computes the expression 
> ```

I'm working on this right now, but it would indeed changed a lot from the current behaviour (which probably isn't bad).

E.g., we allowed

```r
a <- "double_SepWidth = 2 * Sepal.Width"
data_modify(iris, a)
```

This will no longer be valid, because we now _expect_ named arguments, i.e. at least:

```r
a <- "2 * Sepal.Width"
data_modify(iris, double_SepWidth = as_expression(a))
```

strengejacke:
(which will make it a lot easier)

DominiqueMakowski:
To me one of the strength of the easyverse is precisely that it works with strings and not (exclusively) the NSE tidyverse oddity, so we should aim at keep supporting that 100% first,, and then sure we can also make "just work" with expressions passed instead of characters but I wouldn't sweat that such support is limited and not fullproof

strengejacke (commented):
Yes, I also thought about this, and I think I found a robust way to handle expressions, values and expressions as strings.

E.g., these will work:

```r
data_modify(iris, as_expression("sepwid = 2 * Sepal.Width"))
data_modify(iris, as_expression(c("sepwid = 2 * Sepal.Width", "seplen = 5 * Sepal.Length")))

a <- "sepwid = 2 * Sepal.Width"
data_modify(iris, as_expression(a))

a <- c("sepwid = 2 * Sepal.Width", "seplen = 5 * Sepal.Length")
data_modify(iris, as_expression(a))
```

Furthermore, I want to make this work:

```r
data_modify(iris, sepwid = as_expression("2 * Sepal.Width"))
```

everything not masked with `as_expression()` will be used as value, e.g.

```r
data_modify(iris, sepwid = "abc"))

a <- "abc"
data_modify(iris, sepwid = a))
```

strengejacke (commented):
Oh, and of course literal expressions work:

```r
data_modify(iris, sepwid = 2 * Sepal.Width)
data_modify(iris, sepwid = 2 * Sepal.Width, seplen = 5 * Sepal.Length)
```

strengejacke:
Ok, I have to remove some redundant code and fix the grouped_df method, but I think we have it:

``` r
library(datawizard)

data_modify(iris, as_expression("sepwid = 2 * Sepal.Width")) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid
#> 1          5.1         3.5          1.4         0.2  setosa    7.0
#> 2          4.9         3.0          1.4         0.2  setosa    6.0
#> 3          4.7         3.2          1.3         0.2  setosa    6.4
#> 4          4.6         3.1          1.5         0.2  setosa    6.2
#> 5          5.0         3.6          1.4         0.2  setosa    7.2
#> 6          5.4         3.9          1.7         0.4  setosa    7.8

data_modify(
    iris,
    as_expression(c("sepwid = 2 * Sepal.Width", "seplen = 5 * Sepal.Length"))
) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid seplen
#> 1          5.1         3.5          1.4         0.2  setosa    7.0   25.5
#> 2          4.9         3.0          1.4         0.2  setosa    6.0   24.5
#> 3          4.7         3.2          1.3         0.2  setosa    6.4   23.5
#> 4          4.6         3.1          1.5         0.2  setosa    6.2   23.0
#> 5          5.0         3.6          1.4         0.2  setosa    7.2   25.0
#> 6          5.4         3.9          1.7         0.4  setosa    7.8   27.0

data_modify(iris, sepwid = as_expression("2 * Sepal.Width")) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid
#> 1          5.1         3.5          1.4         0.2  setosa    7.0
#> 2          4.9         3.0          1.4         0.2  setosa    6.0
#> 3          4.7         3.2          1.3         0.2  setosa    6.4
#> 4          4.6         3.1          1.5         0.2  setosa    6.2
#> 5          5.0         3.6          1.4         0.2  setosa    7.2
#> 6          5.4         3.9          1.7         0.4  setosa    7.8

data_modify(
    iris,
    sepwid = as_expression("2 * Sepal.Width"),
    seplen = {"5 * Sepal.Length"}
) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid seplen
#> 1          5.1         3.5          1.4         0.2  setosa    7.0   25.5
#> 2          4.9         3.0          1.4         0.2  setosa    6.0   24.5
#> 3          4.7         3.2          1.3         0.2  setosa    6.4   23.5
#> 4          4.6         3.1          1.5         0.2  setosa    6.2   23.0
#> 5          5.0         3.6          1.4         0.2  setosa    7.2   25.0
#> 6          5.4         3.9          1.7         0.4  setosa    7.8   27.0

e <- "sepwid = 2 * Sepal.Width"
data_modify(iris, as_expression(e)) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid
#> 1          5.1         3.5          1.4         0.2  setosa    7.0
#> 2          4.9         3.0          1.4         0.2  setosa    6.0
#> 3          4.7         3.2          1.3         0.2  setosa    6.4
#> 4          4.6         3.1          1.5         0.2  setosa    6.2
#> 5          5.0         3.6          1.4         0.2  setosa    7.2
#> 6          5.4         3.9          1.7         0.4  setosa    7.8

e <- c("sepwid = 2 * Sepal.Width", "seplen = 5 * Sepal.Length")
data_modify(iris, as_expression(e)) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid seplen
#> 1          5.1         3.5          1.4         0.2  setosa    7.0   25.5
#> 2          4.9         3.0          1.4         0.2  setosa    6.0   24.5
#> 3          4.7         3.2          1.3         0.2  setosa    6.4   23.5
#> 4          4.6         3.1          1.5         0.2  setosa    6.2   23.0
#> 5          5.0         3.6          1.4         0.2  setosa    7.2   25.0
#> 6          5.4         3.9          1.7         0.4  setosa    7.8   27.0

e <- "2 * Sepal.Width"
data_modify(iris, sepwid = as_expression(e)) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid
#> 1          5.1         3.5          1.4         0.2  setosa    7.0
#> 2          4.9         3.0          1.4         0.2  setosa    6.0
#> 3          4.7         3.2          1.3         0.2  setosa    6.4
#> 4          4.6         3.1          1.5         0.2  setosa    6.2
#> 5          5.0         3.6          1.4         0.2  setosa    7.2
#> 6          5.4         3.9          1.7         0.4  setosa    7.8

e <- "2 * Sepal.Width"
data_modify(iris, sepwid = as_expression(e), seplen = 5 * Sepal.Length) |> head()
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species sepwid seplen
#> 1          5.1         3.5          1.4         0.2  setosa    7.0   25.5
#> 2          4.9         3.0          1.4         0.2  setosa    6.0   24.5
#> 3          4.7         3.2          1.3         0.2  setosa    6.4   23.5
#> 4          4.6         3.1          1.5         0.2  setosa    6.2   23.0
#> 5          5.0         3.6          1.4         0.2  setosa    7.2   25.0
#> 6          5.4         3.9          1.7         0.4  setosa    7.8   27.0

e <- "2 * Sepal.Width"
f <- "half_petal = 0.5 * Petal.Length"
a <- "string"
num <- 1:5
data_modify(
    iris,
    sepwid = as_expression(e),
    seplen = 5 * Sepal.Length,
    {f},
    new_var = a,
    new_num = num,
    new_var2 = "ho",
    new_num2 = 4:6,
    Sepal.Length = NULL,
    Petal.Length = NULL,
    Sepal.Width = NULL,
    Petal.Width = NULL
) |> head()
#>   Species sepwid seplen half_petal new_var new_num new_var2 new_num2
#> 1  setosa    7.0   25.5       0.70  string       1       ho        4
#> 2  setosa    6.0   24.5       0.70  string       2       ho        5
#> 3  setosa    6.4   23.5       0.65  string       3       ho        6
#> 4  setosa    6.2   23.0       0.75  string       4       ho        4
#> 5  setosa    7.2   25.0       0.70  string       5       ho        5
#> 6  setosa    7.8   27.0       0.85  string       1       ho        6
```

<sup>Created on 2025-04-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
@etiennebacher I think this is ready for review. It would be great if we can include this PR in the next release (which should be soon due to breaking changes in bayestestR)

tests/testthat/test-data_modify.R
```diff
@@ -576,6 +595,163 @@ test_that("data_modify .if/.at arguments", {
...
+test_that("data_modify errors with new expressions", {
+  e <- "sepwid = 2 * Sepal.Widht"
+  expect_error(
+    data_modify(iris, as_expression(e)),
+    regex = "in the first expression"
+  )
+  expect_error(
+    data_modify(iris, as_expression(e)),
+    regex = "Sepal.Widht"
+  )
```
etiennebacher, line 714: Why do we have two `expect_error()` for each case? If there are several messages to test I'd rather have `expect_snapshot(error = TRUE)`.

etiennebacher (changes requested):
Thanks, I'll do my best but it's a big PR so I'll probably need several rounds of review. Also don't forget to update the title.

R/data_modify.R
```diff
@@ -10,21 +10,32 @@
...
+#'   # or
+#'   a <- "center(Sepal.Width)"
+#'   data_modify(iris, Sepal.Width = as_expression(a))
+#'   # or
+#'   a <- "Sepal.Width = center(Sepal.Width)"
+#'   data_modify(iris, as_expression(a))
+#'   ```
+#'   `{}` can be used instead of `as_expression()`, thus, for the above example,
+#'   `Sepal.Width = {a}` or simply `{a}` for the second case would be a valid
+#'   syntax.
```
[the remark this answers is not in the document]
strengejacke: I thought `{}` is for those who find `as_expression()` to long ;-)

tests/testthat/test-data_modify.R
strengejacke: We have just one error, but this error gives different information in the message:
- which expression fails?
- what is the exact typo?

I wanted to check for both.

R/data_modify.R
```diff
@@ -276,6 +278,174 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify =
...
+
+  if (is.null(names(dots)) || !all(nzchar(names(dots)))) {
+    # find which dots are unnamed, check those for expressions
+    if (is.null(names(dots))) {
+      unnamed_dots <- seq_along(dots)
+    } else {
+      unnamed_dots <- which(!nzchar(names(dots)))
+    }
+
+    for (i in rev(unnamed_dots)) {
```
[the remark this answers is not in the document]
strengejacke: Because we possible insert more than one element back into dots, which will corrupt the counter when starting from first element.

Say `i=1`, and we then insert two elements back into `dots[i]`, the loop should continue with `i=3`, not `i=2`. This doesn't matter if we start from the end.

R/data_modify.R
```diff
@@ -276,6 +278,174 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify =
...
+      # copy dot-element and convert to string for manipulation
+      dot_element <- dots[[i]]
+      symbol_string <- insight::safe_deparse(dot_element)
+      # sanity check
+      if (is.null(symbol_string)) next
```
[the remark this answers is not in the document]
strengejacke: I thought I had a case where this happened, not sure exactly which one. When disabling this line, all tests still pass. _Maybe_ this is an old artifact from a previous development stage, which is redundant now.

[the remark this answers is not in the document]
strengejacke: We indeed need it, I added a comment in the code:

```r
data_modify(iris, Species = NULL) |> head()

data_modify(iris, as_expr("Species = NULL")) |> head()
```

R/data_modify.R
```diff
@@ -101,13 +113,16 @@
 #' head(new_efc)
 #'
 #' # works from inside functions
-#' foo <- function(data, z) {
-#'   head(data_modify(data, z))
+#' foo1 <- function(data, ...) {
+#'   head(data_modify(data, ...))
 #' }
-#' foo(iris, "var_a = Sepal.Width / 10")
+#' foo1(iris, SW_fraction = Sepal.Width / 10)
```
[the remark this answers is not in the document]
strengejacke: yes, indeed it works now

strengejacke:
Thanks for the review! All comments are resolved, hence now we have additional features included. Tests should be added for all cases.

NEWS.md
```diff
@@ -14,6 +14,14 @@ BREAKING CHANGES
   requested has changed. Now, for each centrality measure a confidence interval
   is calculated (#617).
 
+* `data_modify()` now always uses values of a vector for a modified or newly
+  created variable, and no longer tries to detect whether a character value
+  possibly contains an expression. To allow expression provided as string (or
+  character vectors), use the helper-function `as_expr()`. Only literal
+  expressions or strings wrapped in `as_expr()` will be evaluated as
+  expressions, everything else will be treated as vector with values for new
+  variables (#605)
```
etiennebacher, line 23:
```suggestion
  variables (#605).
```

R/data_modify.R
```diff
@@ -10,21 +10,29 @@
...
+#'   in the data), or of same length as the data.
+#' - A variable that contains a value to be used. Example:
 #'   ```r
-#'   a <- "2 * Sepal.Width"
-#'   data_modify(iris, a)
+#'   a <- "abc"
+#'   data_modify(iris, var_abc = a) # var_abc contains "abc"
+#'   ```
+#' - An expression can also be provided as string and wrapped in
+#' `as_expr()`. Example:
```
etiennebacher, line 21: It should be mentioned somewhere that `as_expr()` is not a "true" function in the sense that it helps us processing expressions in `data_modify()` but it cannot be used outside and is not exported / documented.

R/data_modify.R
```diff
@@ -276,8 +286,196 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify =
...
+    }
+  }), use.names = FALSE)
+  # now we should have the expression as character string. Next, we
+  # # remove quotes from strings
+  gsub("^\"(.*)\"$", "\\1", symbol_string)
+}
+
+
+.fix_quotes <- function(symbol_string) {
+  # if user uses double-quotes insinde double-quotes, these are escaped by
```
etiennebacher, line 471:
```suggestion
  # if user uses double-quotes inside double-quotes, these are escaped by
```

R/data_modify.R
```diff
@@ -276,8 +286,196 @@ data_modify.grouped_df <- function(data, ..., .if = NULL, .at = NULL, .modify =
...
+  # "\", e.g. data_modify(iris, foo = as_expr("grepl(\"a\", Species)"))
+  # In this case, we have double-backslashes, which need to be removed.
+  # Furthermore, to avoid adding back backslashes, we replace by single quotes
+  gsub("\\", "", gsub("\"", "'", symbol_string, fixed = TRUE), fixed = TRUE)
```
etiennebacher, line 475: Do we need something more robust than this? What happens if one puts single quotes in the regex of the `grepl()` example above?

etiennebacher (changes requested):
Thanks, almost there for me. The implementation is quite complex but the test suite looks complete and I like the fact that one can still build and evaluate expressions from strings.

@DominiqueMakowski does the current behavior look ok to you?

R/data_modify.R
strengejacke: We should then do the same for the select-helpers that are mentioned in the docs for the `select` argument.

R/data_modify.R
strengejacke: Then everything is fine. The problem is that double-quotes are escaped by a backslash, single quotes are not.

strengejacke: I added some tests for all possible combinations

strengejacke:
We can revert the last commit, but wdyt about adding a `by` argument?

R/data_modify.R
etiennebacher: Ok but let's add it here for `as_expr()` since we're revamping the entire function anyway.

strengejacke:
Ok, I think recent comments are also resolved.

etiennebacher (approved):
Alright, thanks for addressing all the comments, I'm happier with the current implementation than the old one. We have 3 weeks to fix CRAN errors so let's merge this now and revisit later if some things must be improved.

strengejacke:
CRAN errors should already be fixed, I think. You can submit any time, but I think you had some reminders/todos in #615 before submission.

PR #606, opened 2025-04-10
`data_read()` no longer coerces Stan models into data frames.

strengejacke:
@DominiqueMakowski This was the issue I was facing when downloading chocominimodel with `data_read()`. It tries to coerce data into a data frame, if possible. And it is possible with stan models. Thus, this PR implements a better check, so you can also download models from internet using `data_read()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 BREAKING CHANGES
 
 * `data_read()` now also returns Bayesian models from packages *brms* and
-  *rstanarm* as original mpdel objects, and no longer coerces them into data
+  *rstanarm* as original model objects, and no longer coerces them into data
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,5 +1,11 @@
 # datawizard (devel)
 
+BREAKING CHANGES
+
+* `data_read()` now also returns Bayesian models from packages *brms* and
+  *rstanarm* as original model objects, and no longer coerces them into data
+  frames.
```
etiennebacher, line 7:
```suggestion
  frames (#606).
```

etiennebacher (approved):
Thanks, can you just fix lints?

PR #607, opened 2025-04-14
Use base-functions for importing R data types

strengejacke:
This PR avoid the usage of the `rio` package for R data types that can be imported using base R functions.

During teaching I realized that when installing easystats packages, users often only install the strictly necessary packages, thus, `rio` is often missing. It's not a big deal to install later, but for base R functions, we don't need this intermediate step, but can have our own import functions.


R/data_read.R
```diff
@@ -300,6 +304,83 @@ data_read <- function(path,
...
+
+  out
+}
+
+
+# check input helper --------------------------------------------------------
+
+# return an URL object if path is a URL
+.check_path_url <- function(path) {
+  url_pattern <- "^(https?|ftp)://(.*)"
```
etiennebacher, line 375: Does this allow `http`? (not sure if we should)

etiennebacher (changes requested):
Thank you, a couple of comments and this also needs an item in NEWS (with the PR number)

strengejacke:
I'm thinking about closing this PR, because when adding more tests for URLs, I realized that I had some troubles loading files from the internet. This didn't even work for `load()`, e.g.

``` r
load(url("https://raw.github.com/easystats/circus/main/data/list_for_testing.rda"))
#> Error in load(url("https://raw.github.com/easystats/circus/main/data/list_for_testing.rda")): the input does not start with a magic number compatible with loading from a connection
```

<sup>Created on 2025-04-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I checked the `rio` code, and it downloads remote files first, then passes them to  `load()` or `readRDS()`, which means it is indeed more effort to copy the functionality from `rio` for base-R file types.

Maybe I leave this open as a draft and get back to it when I have some more time.

R/data_read.R
strengejacke: Yes, it does. Not sure if we should not allow `http://`?

R/data_read.R
```diff
@@ -300,6 +304,83 @@ data_read <- function(path,
   }
   out <- do.call(rio::import, c(rio_args, list(...)))
 
+  # check if loaded file is a data frame, or not (e.g. model objects)
+  file_is_data <- .loaded_file_is_data(out)
+  if (!isFALSE(file_is_data)) {
```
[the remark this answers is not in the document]
strengejacke: No, `.loaded_file_is_data()` returns `FALSE` for non-data objects, and the data frame for data-objects. The latter are _not_ `TRUE`, but they are `!FALSE`.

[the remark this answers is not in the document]
etiennebacher: Then I think `.loaded_file_is_data()` is wrong. Its name implies that it returns `TRUE` or `FALSE` but it actually returns `data.frame` or `FALSE`, which makes the logic hard to understand. 

I think it would be clearer to have:

```r
.get_data_from_loaded_file <- function(out, verbose = TRUE) {
 
# return NULL if file is not data

}
```
and then: 

```r
file_is_data <- nrow(.get_data_from_loaded_file(out)) > 0
if (isTRUE(file_is_data)) {
```

NEWS.md
```diff
@@ -12,6 +12,10 @@ BREAKING CHANGES
 
 CHANGES
 
+* `data_read()` and `data_write()` now rely on base-R functions for files of
+  type `.rds`, `.rda` or `.rdata`, thus, package *rio*  is no longer required
```
etiennebacher, line 16:
```suggestion
  type `.rds`, `.rda` or `.rdata`. Thus, package *rio*  is no longer required
```

R/data_read.R
```diff
@@ -101,6 +102,9 @@ data_read <- function(path,
   out <- switch(file_type,
     txt = ,
     csv = .read_text(path, encoding, verbose, ...),
+    rda = ,
+    rdata = .read_base_rda(path, file_type, verbose, ...),
+    rds = .read_base_rds(path, file_type, verbose, ...),
```
etiennebacher, line 107: Do we need a `file_type` for this one? Isn't it always going to be ".rds"?

tests/testthat/test-data_read.R
```diff
@@ -489,3 +502,19 @@ test_that("data_read - RDA file, model object", {
     expect_s3_class(d, "brmsfit")
   })
 })
+
+
+test_that("data_read - RDS file, from URL", {
+  # works with URL
+  request <- httr::GET("https://raw.github.com/easystats/circus/main/data/model_object.rds")
+  httr::stop_for_status(request)
```
etiennebacher, line 510: This needs a `skip_if_offline()`

etiennebacher (changes requested):

tests/testthat/test-data_read.R
strengejacke: No, we have it in the beginning of the test-file.

strengejacke:
Anything that still needs to be addressed here?

etiennebacher:
No, I think it's good to go, thanks!

etiennebacher (approved):

PR #610, opened 2025-04-14
Rework formatted output of `describe_distribution()`

### LINKED ISSUE #599: Request describe_distribution()
From Andy for his book update:
- [ ] When you use group_by() with describe_distribution() (you're going to tell me not to do this .....) it adds a column to the end .group, c**an the default be to put this column at the start of the tibble, not the end?** 
- [ ] Also, why .group not the name of the variable, and why variable=category rather than the group names? I think it would be really helpufl if the default behaviour was to stick a column at the start of the output, names with the same name as the grouping variable, with rows labelled with the factor levels/categories they represent

Should we add a `by` argument to that function as we now have in more and more ones?

etiennebacher:
Fixes #599.

Here's what I get with a quarto HTML output (on two variables only):

```r
## One group
palmerpenguins::penguins |>
  group_by(species) |>
  describe_distribution(select = c("bill_length_mm", "bill_depth_mm")) |>
  insight::display()

## Two groups
palmerpenguins::penguins |>
  group_by(species, island) |>
  describe_distribution(select = c("bill_length_mm", "bill_depth_mm")) |>
  insight::display()
```

image

@profandyfield @strengejacke @DominiqueMakowski is this what you'd like?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 1.0.2.8
+Version: 1.0.2.9
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0003-1995-6531")),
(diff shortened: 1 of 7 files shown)


strengejacke:
I had nothing particular in mind, but this looks great!

PR #612, opened 2025-04-15
Improvements of output for `describe_distribution()`

### LINKED ISSUE #611: Improvements of output for `describe_distribution()`
Some quick tests ... The specific example in the Chapter I am working looks fine but it doesn't make sense to have the `variable` column if you are only summarizing one variable (as would be  the case if your data are in tidy format) - it's redundant information occupying a column. For example,

```
cloak_tib <- tibble::tibble(
    id = c("Kiara", "Anupama", "Collin", "Ian", "Vanessa", "Darrell", "Tyler", "Steven", "Katheryn", "Kirkpatrick", "Melissa", "Kinaana", "Shajee'a", "Sage", "Jorge", "Conan","Tamara","Man", "Joseph", "Israa", "Kathryn", "Luis", "Devante", "Jerry"),
    cloak = gl(2, 12, labels = c("No cloak", "Cloak")),
    mischief = c(3, 1, 5, 4, 6, 4, 6, 2, 0, 5, 4, 5, 4, 3, 6, 6, 8, 5, 5, 4, 2, 5, 7, 5)
  )

cloak_tib |> 
  group_by(cloak) |> 
  describe_distribution(ci = 0.95) |> 
  display()
```
gives me:

Image

Looks great except for the redundant `Variable` column.

Looking at a `tidy data` example, grouping by two variables:

```
palmerpenguins::penguins |> 
  tidyr::pivot_longer(cols = bill_length_mm:body_mass_g, names_to = "Measure", values_to = "Values") |> 
  group_by(species, Measure) |> 
  describe_distribution(select = "Values", ci = 0.95) |> 
  display()
```

I get :

Image

which look s good, but again the `Variable` column is redundant

Interestingly, if in my tidy data I have a variable called `Variable` (which is quite common I think), it changes the behaviour of the output

```
palmerpenguins::penguins |> 
  tidyr::pivot_longer(cols = bill_length_mm:body_mass_g, names_to = "Variable", values_to = "Values") |> 
  group_by(species, Variable) |> 
  describe_distribution(select = "Values", ci = 0.95) |> 
  display()
```

I get:

Image

The `variable` column is gone (presumably overriden by the one in the data), but the the grouping variables appear in the end two columns, not the first two.

I hope that's useful. I can try more examples if you like!

 

 _Originally posted by @profandyfield in [#599](https://github.com/easystats/datawizard/issues/599#issuecomment-2803184559)_

strengejacke:
Fixes #611


R/data_tabulate.R
```diff
@@ -398,6 +398,11 @@ insight::print_html
 insight::print_md
 
 
+#' @importFrom insight display
+#' @export
+insight::display
```
etiennebacher, line 403: Good idea, can you mention this in NEWS?

R/describe_distribution.R
```diff
@@ -567,3 +603,29 @@ print.parameters_distribution <- function(x, digits = 2, ...) {
...
+    "Min", "Max", "Range", "Trimmed_Mean", "Trimmed", "Mean", "SD", "IQR",
+    "Skewness", "Kurtosis", "n"
+  )
+  invalid_names <- intersect(reserved_names, x)
+
+  if (length(invalid_names) > 0) {
+    # adapt message to show user whether wrong variables appear in grouping or select
+    msg <- switch(type,
+      select = "with `describe_distribution()`: ",
+      "as grouping-variables in `describe_distribution()`: "
```
etiennebacher, line 622:
```suggestion
      "as grouping variables in `describe_distribution()`: "
```

R/describe_distribution.R
```diff
@@ -405,6 +405,9 @@ describe_distribution.data.frame <- function(x,
     verbose = verbose
   )
 
+  # check for reserved variable names
+  .check_for_reserved_names(select)
+
```
etiennebacher, line 410: Isn't this executed too early? For instance:

```r
palmerpenguins::penguins |> 
  data_to_long(cols = bill_length_mm:body_mass_g, names_to = "Variable", values_to = "Values") |> 
  data_group(c("species", "Variable")) |> 
  describe_distribution(select = "Values", ci = 0.95) |> 
  display()
```
would error but the "Variable" column wouldn't be in the output anyway because it is deleted in the `format()` method. So it would be confusing that I get an error for that but if I rename "Variable" to "Measure", the code runs but there's no column "Variable" in the output anyway.

strengejacke: The grouping variables should be shown in the output to identify the rows / groups. It doesn't make sense to remove the grouping variable from the output, or am I missing something?

strengejacke: I think I understand. If `Variable` is to be removed, but in this case it refers to a (grouping) column that should be shown, there is no name conflict.

We then must detect this edge case where a grouping variable equals the name of a column that _could_ be removed (only `Variable`, I think), and if so, check if there is only one unique value, meaning it can be overwritten.

But this would conflict with your other suggestion to keep `Variable` in general. In this case, this check is not too early.

strengejacke:
Just saw, the _methods_ (`display()`, `plot()`, `print_md()`) still live in the _parameters_ package? Should we rename the class here, and have the methods here?

R/describe_distribution.R
etiennebacher: My point was: if I run the following code, I get an error telling me that "Variable" cannot be used as grouping variable because it will conflict:
```r
palmerpenguins::penguins |> 
    data_rename("island", "Variable") |> 
    data_group(c("species", "Variable")) |> 
    describe_distribution(select = "bill_length_mm", ci = 0.95)
```
But if I rename it to say "Measure" and we implement the proposal to drop "Variable" when it has a single value, then it works but I don't have any "Variable" column in the output, so I'm left wondering why there was a conflict in the first place.

This problem goes away if we decide to keep a column "Variable" in the output, no matter the number of unique values it contains.

strengejacke: You don't get any output, because `Variable` is a reserved name, which can't be used as grouping variable. So nothing to wonder about, because you get an error.

strengejacke:
> > Just saw, the methods (display(), plot(), print_md()) still live in the parameters package? Should we rename the class here, and have the methods here?
> 
> Can we move that to a separate issue?

sure.

R/describe_distribution.R
etiennebacher: I get an output if I rename to "Variable" to "Measure", and if we decide to drop the column "Variable" when it contains a single value then I would obtain:
```r
palmerpenguins::penguins |> 
    data_rename("island", "Measure") |> 
    data_group(c("species", "Measure")) |> 
    describe_distribution(select = "bill_length_mm", ci = 0.95)
```
|species   | Measure |  Mean |   SD |  IQR |         95% CI |          Range | Skewness | Kurtosis |   n | n_Missing |
|:---------|:-------:|:-----:|:----:|:----:|:--------------:|:--------------:|:--------:|:--------:|:---:|:---------:|
|Adelie    |  Biscoe | 38.98 | 2.48 | 3.27 | (38.37, 39.74) | (34.50, 45.60) |     0.24 |    -0.14 |  44 |         0 |
|Chinstrap |  Biscoe | 47.50 | 3.08 | 4.30 | (46.95, 48.01) | (40.90, 59.60) |     0.65 |     1.30 | 123 |         1 |
|Gentoo    |  Biscoe | 38.50 | 2.47 | 3.88 | (37.90, 39.03) | (32.10, 44.10) |    -0.15 |    -0.05 |  56 |         0 |
|Adelie    |   Dream | 48.83 | 3.34 | 4.97 | (47.95, 49.48) | (40.90, 58.00) |    -0.09 |     0.04 |  68 |         0 |
|Chinstrap |   Dream | 38.95 | 3.03 | 4.50 | (38.21, 39.69) | (33.50, 46.00) |     0.26 |    -0.43 |  51 |         1 |

Which is confusing because the error message (before I rename the variable) told me there would be a conflict while in this output there isn't any "Variable" column.

strengejacke: Ok, I understand. <sup>(time for coffee to speed up my thinking)</sup>

strengejacke: Let me check the failing tests

etiennebacher: > But then we have it, I'd say. We get an error for name conflicts, and we keep all columns in the output (in this PR)

Oh yes, I didn't see the part about dropping "Variable" had been reverted (coffee needed for me too I guess)

etiennebacher (approved):

PR #613, opened 2025-04-15
Move methods from package parameters to datawizard

strengejacke:


etiennebacher (approved):
I'm a bit confused about whether this affects user in any way, but I think it's worth an item in NEWS regardless. 

Also, does this affect the order of packages for the next release round?

strengejacke:
No, this is a user-invisible change. We will maybe have a short time where we get warnings that methods will be overwritten as long as we also have them in _parameters_, but that's no big deal.

etiennebacher (approved):

PR #629, opened 2025-07-09
Add `table()` and/or `as.table()` methods for `data_tabulate()`

### LINKED ISSUE #628: Add `table()` and/or `as.table()` methods for `data_tabulate()`
These are useful for post-processing the output, especially conducting `chisq.test()` and similar.

strengejacke:
Fixes #628


R/data_tabulate.R
```diff
@@ -465,6 +483,87 @@ as.data.frame.datawizard_tables <- function(x,
...
+  as.table(stats::setNames(x[["N"]], x$Value))
+}
+
+#' @export
+as.table.datawizard_tables <- function(x, remove_na = TRUE, ...) {
+  out <- lapply(x, as.table.datawizard_table, remove_na = remove_na, ...)
+  # if only one table is returned, "unlist"
+  if (length(out) == 1) {
+    out <- out[[1]]
+  }
```
etiennebacher, line 508: Since the user passes a list as input, it seems weird to me to return something else than a list. Is it a problem to let the user handle the case of length 1 output?

R/data_tabulate.R
```diff
@@ -465,6 +483,87 @@ as.data.frame.datawizard_tables <- function(x,
...
+}
+
+#' @export
+as.table.datawizard_crosstab <- function(x, remove_na = TRUE, ...) {
+  # data frame is in a list, so we extract the first element
+  if (is.data.frame(x)) {
+    out <- x
+  } else {
+    out <- x[[1]]
+  }
```
etiennebacher, line 519: There's a similar but slightly different logic at the top of `as.table.datawizard_table()`, can you harmonize them?

R/data_tabulate.R
```diff
@@ -465,6 +483,87 @@ as.data.frame.datawizard_tables <- function(x,
...
+  as.table(as.matrix(out))
+}
+
+#' @export
+as.table.datawizard_crosstabs <- function(x, remove_na = TRUE, ...) {
+  out <- lapply(x, as.table.datawizard_crosstab, remove_na = remove_na, ...)
+  # if only one table is returned, "unlist"
+  if (length(out) == 1) {
+    out <- out[[1]]
+  }
```
etiennebacher, line 548: Same comment as above

tests/testthat/test-data_tabulate.R
```diff
@@ -508,3 +508,41 @@ test_that("data_tabulate, as.data.frame, cross tables with total N", {
   expect_identical(nrow(out$table[[1]]), 5L)
   expect_identical(out$table[[1]]$cyl, c("4", "6", "8", NA, "Total"))
 })
+
+
+# table methods -----------------------------
+
+test_that("data_tabulate, table methods", {
```
etiennebacher, line 515: It would be much easier to review snapshot tests (also to check potential changes in the future)

R/data_tabulate.R
```diff
@@ -16,7 +16,10 @@
 #' @param name Optional character string, which includes the name that is used
 #' for printing.
 #' @param remove_na Logical, if `FALSE`, missing values are included in the
-#' frequency or crosstable, else missing values are omitted.
+#' frequency or crosstable, else missing values are omitted. Note that the
+#' default for the `as.table()` method is `remove_na = TRUE`, so that missing
+#' values are not included in the returned table, which makes more sense for
+#' post-processing of the table, e.g. using `chisq.test()`.
```
etiennebacher, line 22: Sounds weird to have different default values, but maybe the main usecase for calling `as.table()` is to run a test right after. Would it make sense to add a message "Dropping NAs" by default in `as.table()`?

etiennebacher (changes requested):
Thanks, a few comments.

tests/testthat/test-data_tabulate.R
strengejacke: I can also add some snapshot tests, but would still keep the existing ones as well.

R/data_tabulate.R
strengejacke: Ok, then I'd also add a `verbose` argument

R/data_tabulate.R
strengejacke: I think `as.table()` should return a table, when the input also is only "one frequency table", and if the user computed *multiple* frequency tables, then `as.table()` should return a list of tables. The user usually doesn't know that the default-method of `data_tabulate()` returns a data frame, while the `.data.frame` method returns a list with one data frame (or more data frames for multiple tables).

strengejacke:
Ok, done. Default is always a list, `simplify` simplifies one-table inputs directly into a table object.

mattansb:
Very nice!

``` r
mtcars |> 
  datawizard::data_tabulate("cyl") |> 
  as.table(simplify = TRUE) |> 
  effectsize::cohens_w()
#> Removing NA values from frequency table.
#> Cohen's w |       95% CI
#> ------------------------
#> 0.27      | [0.00, 1.41]
#> 
#> - One-sided CIs: upper bound fixed at [1.41~].

mtcars |> 
  datawizard::data_tabulate("cyl", by = "gear") |> 
  as.table(simplify = TRUE) |> 
  effectsize::tschuprows_t()
#> Removing NA values from frequency table.
#> Tschuprow's T (adj.) |       95% CI
#> -----------------------------------
#> 0.48                 | [0.07, 1.00]
#> 
#> - One-sided CIs: upper bound fixed at [1.00].
```

<sup>Created on 2025-07-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

But shouldn't `as.table(<datawizard_crosstab>)` never return a list, as it's only one table?

PR #634, opened 2025-07-16
In `as.table()`, only warn about removing `NA`s when there are `NA`s

### LINKED ISSUE #633: Only warn about removing NAs where there are NAs
Can we pop this warnings only if `NA`s are actually to be removed?

https://github.com/easystats/datawizard/blob/5632d4dcfa917197219f230a0c30373dc8260fed/R/data_tabulate.R#L508-L510

https://github.com/easystats/datawizard/blob/5632d4dcfa917197219f230a0c30373dc8260fed/R/data_tabulate.R#L527-L529

https://github.com/easystats/datawizard/blob/5632d4dcfa917197219f230a0c30373dc8260fed/R/data_tabulate.R#L567-L569

https://github.com/easystats/datawizard/blob/5632d4dcfa917197219f230a0c30373dc8260fed/R/data_tabulate.R#L590-L592

strengejacke:
Fixes #633

### DIFF
--- a/tests/testthat/test-data_tabulate.R
+++ b/tests/testthat/test-data_tabulate.R
@@ -586,6 +586,7 @@ test_that("data_tabulate, table methods, only warn if necessary", {
   # single variable
   expect_message(as.table(data_tabulate(efc$c172code)))
   expect_silent(as.table(data_tabulate(efc$c172code, remove_na = TRUE)))
+  expect_silent(as.table(data_tabulate(efc$c172code), remove_na = FALSE))
   expect_silent(as.table(data_tabulate(efc$c172code), verbose = FALSE))
 
   # cross table
@@ -594,6 +595,7 @@ test_that("data_tabulate, table methods, only warn if necessary", {
     regex = "Removing NA values"
   )
   expect_silent(as.table(data_tabulate(efc, "c172code", by = "e42dep", remove_na = TRUE)))
+  expect_silent(as.table(data_tabulate(efc, "c172code", by = "e42dep"), remove_na = FALSE))
   expect_silent(as.table(data_tabulate(efc, "c172code", by = "e42dep"), verbose = FALSE))
 
   # no missings


R/data_tabulate.R
```diff
@@ -620,6 +627,34 @@ as.table.datawizard_crosstabs <- function(x, remove_na = TRUE, simplify = FALSE,
   isTRUE(attributes(x)$grouped_df)
 }
 
+.check_table_na <- function(x) {
+  # check if any table has NA values - the column "Value" contains the value
+  # "NA", and the column "N" contains the frequency of this value.
+  any(vapply(x, function(i) any(i$N[is.na(i$Value)] > 0), logical(1)))
```
etiennebacher, line 633: Same

R/data_tabulate.R
```diff
@@ -620,6 +627,34 @@ as.table.datawizard_crosstabs <- function(x, remove_na = TRUE, simplify = FALSE,
...
+}
+
+.check_xtable_na <- function(x) {
+  any(vapply(x, function(i) {
+    browser()
```
etiennebacher, line 638:
```suggestion
```

R/data_tabulate.R
```diff
@@ -620,6 +627,34 @@ as.table.datawizard_crosstabs <- function(x, remove_na = TRUE, simplify = FALSE,
...
+    # check for NA columns and rows
+    if (!is.null(i[["NA"]])) {
+      has_na <- any(i[["NA"]] > 0)
+    }
+    if ("NA" %in% row_names) {
+      # we need "as.data.frame()" for grouped df
+      if (isTRUE(attributes(i)$grouped_df)) {
+        i$Group <- NULL
+      }
+      has_na <- has_na | any(as.vector(as.data.frame(i[row_names == "NA", -1])) > 0)
```
etiennebacher, line 652: Can `as.vector(as.data.frame())` be simplified to `as.vector()` only ? (I haven't tested)

R/data_tabulate.R
```diff
@@ -504,7 +504,9 @@ as.table.datawizard_table <- function(x, remove_na = TRUE, simplify = FALSE, ver
   if (!is.data.frame(x)) {
     x <- x[[1]]
   }
-  if (remove_na) {
+  # check if any table has NA values - the column "Value" contains the value
+  # "NA", and the column "N" contains the frequency of this value.
+  if (remove_na && any(x$N[is.na(x$Value)] > 0)) {
```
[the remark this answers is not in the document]
strengejacke: No, because we don't have `NA` in the table, only counts of `NA`s. If no NA, `x$Value` is still `NA`, but `x$N = 0`.

R/data_tabulate.R
strengejacke: See above

R/data_tabulate.R
strengejacke: No, because for grouped_df, if a tidyverse package is loaded, simplification with `as.vector()` doesn't work, so we need a "real" data frame first.

etiennebacher (approved):
Thank you, LGTM. Feel free to merge if you're happy with the discussion in #633

PR #636, opened 2025-08-10
Fix issue with nested demean

### LINKED ISSUE #635: Nested demean
Nesting for `datawizard::demean` doesn't seem to exactly work per the instructions in certain cases:

I made up this dataframe:
```
testdf <- data.frame(
    roman = c("I", "I", "I", "I", "II", "II", "II", "II"),
    alphabet = c("a", "a", "b", "b", "a", "b", "a", "b"),
    val1 = c(1, 2, 3, 4, 5, 6, 7, 8),
    val2 = c(1, 2, 3, 4, 5, 6, 7, 8),
    val3 = c(1, 2, 3, 4, 5, 6, 7, 8)
)
```

Now I do:
`datawizard::demean(testdf, select = c("val1", "val2"), by = "roman/alphabet")`
I get columns `val1_roman_between' and `val1_roman.1_between` which are the same. The `.1_between` seems like an artifact.

Second, it doesn't work if I select more than two variables to demean. For example,
`datawizard::demean(testdf, select = c("val1", "val2", "val3"), by = "roman/alphabet")`
I get the error:
```Error in names(out) <- paste0(select, "_", by) : 
  'names' attribute [3] must be the same length as the vector [2]
```
I think it doesn't work if the number of variables to demean exceeds the number of nesting levels.

Finally, the documentation says: "Brincks et al. (2017) have suggested an algorithm to center variables for nested designs, which is implemented in demean()." But that study looks at different ways to demean the level 2 coefficient in a three-level model without offering one particular choice. So the documentation could clarify that datawizard::demean centers each intermediate variable relative to the mean of the group above it, but just reports the mean of the top level---as far as I can tell. Thank you.

mattansb:
Closes #635

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# datawizard 1.2.0
+# datawizard 1.2.0.xx
 
 BUG FIXES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # datawizard 1.2.0
 
+BUG FIXES
+
+* Fixed an issue when `demean()`ing nested structures (#635).
+
+# datawizard 1.2.0
+
 BREAKING CHANGES
 
 * The following deprecated arguments have been removed (#603):
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# datawizard 1.2.0.xx
+# datawizard (development)
 
 BUG FIXES
 
--- a/tests/testthat/test-demean.R
+++ b/tests/testthat/test-demean.R
@@ -266,20 +266,26 @@ test_that("demean for nested designs (by > 1), nested = TRUE", {
     "val1_within", "val2_within", "val3_within"
   ))
 
-  expect_equal(as.vector(out$val1_within),
-               c(-0.5, 0.5, -0.5, 0.5, -1, -1, 1, 1))
+  expect_equal(
+    as.vector(out$val1_within),
+    c(-0.5, 0.5, -0.5, 0.5, -1, -1, 1, 1)
+  )
   expect_equal(out$val1_within, out$val2_within)
   expect_equal(out$val1_within, out$val3_within)
 
-  expect_equal(as.vector(out$val1_roman_between),
-               c(2.5, 2.5, 2.5, 2.5, 6.5, 6.5, 6.5, 6.5))
+  expect_equal(
+    as.vector(out$val1_roman_between),
+    c(2.5, 2.5, 2.5, 2.5, 6.5, 6.5, 6.5, 6.5)
+  )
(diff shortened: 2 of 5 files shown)


tests/testthat/test-demean.R
```diff
@@ -245,4 +245,24 @@ test_that("demean for nested designs (by > 1), nested = TRUE", {
...
+  out <- datawizard::demean(testdf,
+    select = c("val1", "val2", "val3"),
+    by = "roman/alphabet", append = FALSE
+  )
+
+  expect_in(names(out), c(
+    "val1_roman_between", "val1_alphabet_between", "val2_roman_between",
+    "val2_alphabet_between", "val3_roman_between", "val3_alphabet_between",
+    "val1_within", "val2_within", "val3_within"
+  ))
```
etiennebacher, line 267: Can this be `expect_named()` instead?

tests/testthat/test-demean.R
```diff
@@ -245,4 +245,24 @@ test_that("demean for nested designs (by > 1), nested = TRUE", {
...
   )
+
+  # Following #635
+  testdf <- data.frame(
+    roman = c("I", "I", "I", "I", "II", "II", "II", "II"),
+    alphabet = c("a", "a", "b", "b", "a", "b", "a", "b"),
+    val1 = c(1, 2, 3, 4, 5, 6, 7, 8),
+    val2 = c(1, 2, 3, 4, 5, 6, 7, 8),
+    val3 = c(1, 2, 3, 4, 5, 6, 7, 8)
+  )
```
etiennebacher, line 256: Can we also test the values and not only the column names?

etiennebacher (changes requested):
Thanks, a couple of comments on tests.

Also, can you update NEWS and the version in DESCRIPTION?

etiennebacher (commented):
Also, it seems most linting failures can be fixed by using `|>` instead of `%>%` in the test file. Can you fix those?

mattansb:
`|>` is not yet supported in `datawizard` (requires R>4.0, but the native pipe was introduced in 4.1).

Made changed to the tests as you requested.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # datawizard 1.2.0
```
etiennebacher, line 1: Can you fix the version number in NEWS?

There's also some spelling and styling to fix and then it's good to go.

mattansb: Added `1.2.0.xx`

etiennebacher: Thanks, just styling and spelling left ;)

mattansb:
Styler is done.
Spelling is a false positive.

etiennebacher:
Thanks, I just updated the wordlist so that we don't have the CI error anymore.

I'll merge when everything is green.

PR #641, opened 2025-09-01
Fix snapshot test failures due to insight 1.4.1

etiennebacher:


etiennebacher:
@strengejacke you mentioned expected failures on macos-oldrel when using the latest `insight`: https://github.com/easystats/workflows/pull/82#issuecomment-3241020706

But here it's failing on R release (4.5.1) on macos because of errors when installing `Matrix`. Is this expected too?

strengejacke:
_insight_ is failing on CRAN for macos oldrel. If you depend on _insight 1.4.1_, the next datawizard release on CRAN will also fail to pass/built on that os. It's not related to the workflows, I was just referring to the package dependency and CRAN checks.

strengejacke:
> But here it's failing on R release (4.5.1) on macos because of errors when installing Matrix. Is this expected too?

No.

strengejacke:
See also my comment here:
https://github.com/easystats/datawizard/pull/640#issuecomment-3241278829

PR #651, opened 2025-09-08
Performance issues with `data_filter()` when tidyverse is loaded?

### LINKED ISSUE #650: Performance issues with `data_filter()` when tidyverse is loaded?
I realized that a simple code:

```r
    x <- data_group(dakMerge[1:50, ], "ID")
    data_filter(x, sum(!is.na(psmu_wg_rs)) >= 3)   
```

takes ~0.23 seconds on average when a) tidyverse is not loaded or b) we explicitly convert the input in `data_filter.grouped_df()` to a data frame.

Else - I don't know why - the same codes takes ~6 seconds on average.

I profiled the code, and indeed, it seems to be related to some tidyverse-stuff going on:

Image

strengejacke:
Fixes #650

### DIFF
--- a/R/data_match.R
+++ b/R/data_match.R
@@ -183,6 +183,15 @@ data_filter <- function(x, ...) {
 #' @export
 data_filter.data.frame <- function(x, ...) {
   out <- x
+
+  # convert tibble to data.frame
+  if (inherits(data, "tbl_df")) {
+    out <- as.data.frame(out, stringsAsFactors = FALSE)
+    tbl_input <- TRUE
+  } else {
+    tbl_input <- FALSE
+  }
+
   dots <- match.call(expand.dots = FALSE)[["..."]]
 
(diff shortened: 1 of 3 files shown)


R/data_match.R
```diff
@@ -183,6 +183,15 @@ data_filter <- function(x, ...) {
...
 data_filter.data.frame <- function(x, ...) {
   out <- x
+
+  # convert tibble to data.frame
+  if (inherits(data, "tbl_df")) {
+    out <- as.data.frame(out, stringsAsFactors = FALSE)
+    tbl_input <- TRUE
+  } else {
+    tbl_input <- FALSE
+  }
```
etiennebacher, line 193: I'll review more properly later but does this also strip other attributes from the `data`? For instance if there's a `geometry` column added by `sf`

strengejacke: (gitblame says the code is from you 😎 )

PR #656, opened 2025-09-18
Extracting proportions from `data_tabulate()`

### LINKED ISSUE #655: Extracting proportions from `data_tabulate()`
Is there a mechanism for extracting the proportions from a `data_tabulate()` object? This would be useful for report writing/automating reports.

For example, I have a contingency table to which I add column proportions:

``` r
dance_tib <- discovr::cat_dance
prop_dance <- dance_tib |>
  datawizard::data_tabulate(select = "training", by = "dance", proportions = "column", remove_na = TRUE)

# display the table
prop_dance |> 
  insight::display()
```

| training  |          No | Yes        | Total |
|:----------|------------:|:-----------|------:|
| Food      |   10 (8.1%) | 28 (36.8%) |    38 |
| Affection | 114 (91.9%) | 48 (63.2%) |   162 |
|           |             |            |       |
| Total     |         124 | 76         |   200 |

<sup>Created on 2025-09-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Let's say I want to report these proportions in my automated quarto doc as either the raw proportions or percentages, it'd be handy to extract them directly from `prop_dance`. I've tried `as.table()` and `as.data.frame()` but these return only the frequencies and `prop_dance` itself appears to be a printed object so stores no info. Is there an `easystats` function that will extract the proportions for me? Better still, could `data_tabulate()` be changed so that it creates an object that stores the info in the table and `format_table()`/`display()`/`print_md()` etc are the mechanisms through which the information is then formatted/printed?

strengejacke:
Fixes #655

### DIFF
--- a/R/data_xtabulate.R
+++ b/R/data_xtabulate.R
@@ -80,9 +80,40 @@
   attr(out, "proportions") <- proportions
   attr(out, "varname") <- obj_name
   attr(out, "grouped_df") <- !is.null(group_variable)
+  attr(out, "prop_table") <- .prop_table(out)
 
   class(out) <- c("datawizard_crosstab", "data.frame")
 
+  out
+}
+
+
+.prop_table <- function(x) {
+  proportions <- attributes(x)$proportions
+  out <- NULL
+  if (!is.null(proportions)) {
+    # find numeric columns, only for these we need row/column sums
+    numeric_columns <- vapply(x, is.numeric, logical(1))
+    total_n <- attributes(x)$total_n
+
+    out <- switch(
+      proportions,
+      row = lapply(seq_len(nrow(x)), function(i) {
+        row_sum <- sum(x[i, numeric_columns], na.rm = TRUE)
+        x[i, numeric_columns] / row_sum
+      }),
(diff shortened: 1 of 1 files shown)


strengejacke:
@etiennebacher Maybe we can move the code to create the crosstable into an own function, and then simplify (and only do) the formatting in `format()`. Then we have the proportions accessible as attribute and can maybe add an `as.prop.table()` method or something.

etiennebacher:
Sorry, I forgot about this one. I'll make a CRAN release this week so it would be nice to have this PR in it too.

> Maybe we can move the code to create the crosstable into an own function, and then simplify (and only do) the formatting in format(). Then we have the proportions accessible as attribute and can maybe add an as.prop.table() method or something.

Sounds good to me but isn't this already the case since the crosstable is built in `.crosstable()`?

strengejacke:
The proportions are currently added in the `format()` method. In this PR, I wrote a `.prop_table()` function that calculates proportions and adds them as attributes, so we can easily extract then. We should built upon this function in `format()`. to avoid duplicated code .

PR #657, opened 2025-09-19
Improve performance by using `tryCatch()` instead of `try()`

strengejacke:
Both when evaluation succeeds or fails, `tryCatch()` is faster than `try()`. This can be beneficial when we have repeated calls to function, e.g. for grouped data frames.

``` r
microbenchmark::microbenchmark(
  try(x <- 1, silent = TRUE),
  tryCatch(x <- 1, error = function(e) NULL),
  times = 1000
)
#> Unit: microseconds
#>                                        expr   min    lq     mean median    uq
#>                  try(x <- 1, silent = TRUE) 6.100 6.401 7.846637  6.601 7.101
#>  tryCatch(x <- 1, error = function(e) NULL) 5.701 6.001 7.157679  6.201 6.700
#>      max neval cld
#>  220.501  1000  a 
#>   58.600  1000   b
microbenchmark::microbenchmark(
  try(x <- a, silent = TRUE),
  tryCatch(x <- a, error = function(e) NULL),
  times = 1000
)
#> Unit: microseconds
#>                                        expr    min      lq      mean  median
#>                  try(x <- a, silent = TRUE) 87.202 92.2010 105.88034 99.3505
#>  tryCatch(x <- a, error = function(e) NULL) 29.201 31.4015  35.81722 33.5015
#>        uq     max neval cld
#>  110.4015 352.901  1000  a 
#>   36.3010 172.801  1000   b
```

<sup>Created on 2025-09-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In a recent code, I had speed improvement of ~4 seconds - however, the major bottleneck when _datawizard_ deals with grouped df is that we just call the `.data.frame` method repeatedly. In the long run, we might think about writing loops inside the `.data.frame` method for grouped df's instead of repeatedly calling it.

I found this for `data_filter()`, where the profiling showed that `.dynEval()` slows down the filtering:

```r
  for (i in seq_along(dots)) {
    # only proceed when result is still valid
    if (!is.null(out)) {
      symbol <- dots[[i]]
      # evaluate, we may have a variable with filter expression
      eval_symbol <- .dynEval(symbol, ifnotfound = NULL)
      # validation check: is variable named like a function?
      if (is.function(eval_symbol)) {
        eval_symbol <- .dynGet(symbol, ifnotfound = NULL)
      }
...
```

Maybe we can re-write the code so that dots are evaluated first, and then we filter on subsets of grouped df in the above code?!? But that's not part of this PR.

My idea, roughly:

```r
  for (i in seq_along(dots)) {
    # only proceed when result is still valid
    if (!is.null(out)) {
      symbol <- dots[[i]]
      # evaluate, we may have a variable with filter expression
      eval_symbol <- .dynEval(symbol, ifnotfound = NULL)

# after evaluating symbol, we now split a grouped data frame
# into subsets and do not repeatedly
# evaluate the dots for each subset of the grouped df.
```

### DIFF
--- a/R/select_nse.R
+++ b/R/select_nse.R
@@ -236,6 +236,10 @@
 # it is a select helper that we grab from the error message.
 
 .select_symbol <- function(data, x, ignore_case, regex, verbose, ifnotfound) {
+  # We use `tryCatch()` instead of `try()` here, because for grouped data frame
+  # methods, `.dynEval()` can be called many times. Since `tryCatch()` is minimal
+  # faster than `try()`, we get a performance "boost" of some seconds for large
+  # data frames with many groups (see https://github.com/easystats/datawizard/pull/657/)
   try_eval <- tryCatch(eval(x), error = function(e) NULL)
   x_dep <- insight::safe_deparse(x)
   is_select_helper <- FALSE
@@ -632,6 +636,10 @@
       n <- n - 1L
       env <- sys.frame(n)
     }
+    # We use `tryCatch()` instead of `try()` here, because for grouped data frame
+    # methods, `.dynEval()` can be called many times. Since `tryCatch()` is minimal
+    # faster than `try()`, we get a performance "boost" of some seconds for large
+    # data frames with many groups (see https://github.com/easystats/datawizard/pull/657/)
     r <- tryCatch(eval(str2lang(x), envir = env), error = function(e) NULL)
     if (!is.null(r)) {
       return(r)
--- a/R/select_nse.R
+++ b/R/select_nse.R
@@ -236,7 +236,7 @@
 # it is a select helper that we grab from the error message.
 
 .select_symbol <- function(data, x, ignore_case, regex, verbose, ifnotfound) {
-  try_eval <- try(eval(x), silent = TRUE)
+  try_eval <- tryCatch(eval(x), error = function(e) NULL)
   x_dep <- insight::safe_deparse(x)
   is_select_helper <- FALSE
   out <- NULL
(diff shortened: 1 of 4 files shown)


strengejacke:
I see that we cannot replace all instances, because sometimes, returning `NULL` is a valid value. I'll check test results and see where we can safely replace `try()` with `tryCatch()`.

strengejacke:
Data is confidential, but I can send by mail, if you're interested.

etiennebacher:
> Data is confidential, but I can send by mail, if you're interested.

I don't think this is necessary, I can do fake data with many groups. I'll take a look this weekend.

R/select_nse.R
```diff
@@ -236,7 +236,7 @@
 # it is a select helper that we grab from the error message.
 
 .select_symbol <- function(data, x, ignore_case, regex, verbose, ifnotfound) {
-  try_eval <- try(eval(x), silent = TRUE)
+  try_eval <- tryCatch(eval(x), error = function(e) NULL)
```
etiennebacher, line 239: Can you add a comment to say we don't want `try(silent = TRUE)` with a link to this PR?

R/select_nse.R
```diff
@@ -632,8 +632,8 @@
       n <- n - 1L
       env <- sys.frame(n)
     }
-    r <- try(eval(str2lang(x), envir = env), silent = TRUE)
-    if (!inherits(r, "try-error") && !is.null(r)) {
+    r <- tryCatch(eval(str2lang(x), envir = env), error = function(e) NULL)
+    if (!is.null(r)) {
```
etiennebacher, line 636: Same

etiennebacher (approved):
LGTM, thanks, just a minor comment.

Did you check that the performance improvements are still present only with these changes? If not, I can check and merge tonight.

strengejacke:
> Did you check that the performance improvements are still present only with these changes? If not, I can check and merge later today.

Yes, for me there's still a performance boost of ~4 seconds (roughly 22 vs 26 seconds), but I tried to include one more change. Let's see if tests pass.

PR #660, opened 2025-10-07
Format the package with Air

etiennebacher:
@easystats/maintainers Steps to reproduce this kind of PR in the other `easystats` repos:

1. `usethis::use_air(vscode = TRUE/FALSE)` depending on whether you use VS Code
2. put this in the new `air.toml`:
```
[format]
line-width = 80
indent-width = 2
indent-style = "space"
line-ending = "lf"
persistent-line-breaks = true
skip = ["tribble"]
```
(I think the line width and indent can be changed according to your preferences since they will be picked up by Air in CI and when formatting on save anyway)
3. `air format .`
4. after merging the PR, add the corresponding commit in `.git-blame-ignore-revs` (you may need to create this file first): [medium.com/@ramunarasinga/git-blame-ignore-revs-to-ignore-bulk-formatting-changes-f20ac23e6155](https://medium.com/@ramunarasinga/git-blame-ignore-revs-to-ignore-bulk-formatting-changes-f20ac23e6155)

### DIFF
--- a/air.toml
+++ b/air.toml
@@ -1,0 +1,7 @@
+[format]
+line-width = 80
+indent-width = 2
+indent-style = "space"
+line-ending = "lf"
+persistent-line-breaks = true
+skip = ["tribble"]


etiennebacher:
@strengejacke the only thing to review are the Air settings, in case you have some preferences.

PR #673, opened 2026-03-10
Allow expressions in `data_summary()` that return > 1 column summaries

strengejacke:
Revision


etiennebacher (commented):
I wonder if this is necessary, it's always the debate about mimicking `tidyverse` or not. The motivating example is just one case, and I personally never use `reframe()`.

That said, if you have cases that would benefit from it, that's ok for me but I have a few comments below. I also wonder if `strict` is clear enough, or if something like `allow_multiple` is clearer, but I'll leave that up to you.

etiennebacher (changes requested):
Looks good to me but I'd like to wait for @mattansb's opinion before merging.

etiennebacher (approved):
Looks good, just a suggestion to reorganize a bit the param description.

Thanks!

PR #675, opened 2026-04-08
`data_read()` can encrypt/decrypt R data files

strengejacke:


strengejacke:
@gemini-code-assist review

etiennebacher (commented):
I don't feel knowledgeable enough to review this PR. I've never had to interact with password-protected datafiles and I don't know the best practices in encryption/decryption.

Also, I don't see any feature request to handle encrypted files in `haven` or `readr`, so I think this is rare enough to let users explicitly handle it rather than doing some internal magic for them. I don't think the potential usability gain is worth the extra maintenance effort and risk.

strengejacke:
I used this script to encrypt/decrypt data for a longer time now, and it's quite simple. I don't think the maintenance effort is very high. It's quite convenient, because data encryption is increasingly important (and awareness is raising). My colleagues use a password-protected zip-folder, which is of course annoying to handle. Same for loading and running the R script. I think a direct implementation is very practical.

etiennebacher:
My concern is that this becomes the top result when people search for "R write password protected rda" while I'm not confident this follows best practices in security. For example, the current implementation uses:

-  `openssl::sha256()` but why not `md5()`, or `sha512()`, or ...
-  `openssl::aes_cbc_decrypt()`, but why not `aes_ctr_decrypt()`, or `aes_gcm_decrypt()`, or ...

My point is that there are various key components that we're just hiding from the user here and where we assume we make the right choice for them, but I'd rather have them being explicit about the functions they're using to secure data. The implementation is 4-5 lines for each so I don't think it's too hard for users (we could also add a line in docs to redirect them to openssl to encrypt/decrypt data).

Curious about @easystats/core-team opinions here

strengejacke:
Ok, I see the point. So the minimum, if we decide on integrating such a feature, would be to use more up-to-date encryption standards?

etiennebacher:
Yes but also give the user more choice to use the method they want. But my point was also that since the implementation is very simple, I think we'd just be better off documenting how to do this and redirect to `openssl` or other encryption/decryption-focused packages without having to maintain something to do that ourselves.

strengejacke:
After a little chat with Gemini, I learned that sha512 is the most secure hash key generation, and gcm-encryption is the most secure encryption method, so I updated the code here. I think this is in line with the help file from openssl.

>  I think we'd just better off documenting how to do this and redirect to openssl or other encryption/decryption-focused packages

But part of the easystats philosophy is to keep these confusing stuff away from users. Most of the people around me who work with R can use basics, maybe a bit of programming, but would not know how to encrypt/decrypt data (or other stuff). That's why I think it's convenient to have it here, and I don't think it will lead to "bad practices" in terms of "bad encryption" or security. Rather, if there's no easy-to-understand way how to encrypt files, it's often not done at all. So we would definitely have an improvement here for casual users, that's my impression.

R/data_read.R
```diff
@@ -77,12 +81,22 @@
...
 #'
+#' @section Data encryption:
+#' `data_read()` and `data_write()` support data encryption for R file formats
+#' (`.rds`, `.rda` and `.rdata`). To encrypt a file, provide a password to the
+#' `password` argument in `data_write()`. To decrypt the file, provide the same
+#' password to `data_read()`. The encryption is based on the **openssl** package
+#' and uses the AES-GCM algorithm (see `?openssl::aes_gcm_encrypt`) with a
+#' 256-bit key (see `?openssl::sha256`). **Warning:** Do not lose your
+#' `password`, else you will not be able to decrypt the data again!
+#'
```
etiennebacher, line 103: sha512 instead of sha256?

Also, this should mention that the decryption method only works when the file was encrypted with `data_write()` or in another function but still using openssl::sha512

R/data_read.R
```diff
@@ -522,3 +541,39 @@ data_read <- function(
...
+    data <- .decrypt_data(data, password)
+  }
+  data
+}
+
+.decrypt_data <- function(data, password = NULL) {
+  insight::check_if_installed("openssl", "for data decryption")
+  # it is important to remember the phrase! else, you cannot decrypt the data
+  passphrase <- charToRaw(password)
+  key <- openssl::sha256(passphrase)
```
etiennebacher, line 571: Should be sha512?

R/data_read.R
```diff
@@ -522,3 +541,39 @@ data_read <- function(
...
+
+.data_decryption <- function(data, password = NULL) {
+  # check if data should be decrypted
+  if (!is.null(password)) {
+    # password needs to be a character string
+    if (!is.character(password)) {
+      insight::format_error(
+        "The `password` argument must be a character string."
+      )
+    }
```
etiennebacher: This is different from the check in `.data_encryption()`, worth harmonizing by extracting a helper `.check_password()`.

R/data_write.R
```diff
@@ -341,3 +379,39 @@ data_write <- function(
...
+
+.data_encryption <- function(data, password = NULL) {
+  # check if data should be encrypted
+  if (!is.null(password)) {
+    # password needs to be a character string
+    if (
+      !is.character(password) || length(password) != 1L || !nzchar(password)
+    ) {
+      insight::format_error(
+        "The password must be a signle non-empty character string."
```
etiennebacher: Typo in "single"

tests/testthat/test-data_write.R
```diff
@@ -14,6 +14,64 @@ d <- data_filter(efc, 1:5)
 d$e42dep <- droplevels(d$e42dep)
 
 
+# data encryption with rds ------------------
```
etiennebacher, line 17: There should be several checks to ensure password is of the expected format in both functions.

tests/testthat/test-data_write.R
```diff
@@ -14,6 +14,64 @@ d <- data_filter(efc, 1:5)
...
+    expect_named(d2, "encrypted")
+
+    # password, returns decrypted data frame
+    d2 <- data_read(tmp, password = "test")
+    expect_identical(d, d2)
+
+    # wrong password
+    expect_error(data_read(tmp, password = "text"), "File does not appear")
+
+    # not encrypzed
```
etiennebacher: typo

tests/testthat/test-data_write.R
```diff
@@ -14,6 +14,64 @@ d <- data_filter(efc, 1:5)
...
+    data_write(d, tmp)
+    expect_error(data_read(tmp, password = "test"), "File does not appear")
+  })
```
etiennebacher, line 78: There should be some checks that encrypting with another function and decrypting with data_read() fails.

tests/testthat/test-data_write.R
```diff
@@ -14,6 +14,64 @@ d <- data_filter(efc, 1:5)
...
+
+test_that("data_write, encrypting rds files", {
+  skip_if_not_installed("withr")
+  skip_if_not_installed("openssl")
+  withr::with_tempfile("tmp", fileext = ".rds", code = {
+    expect_warning(data_write(d, tmp, password = "test"), "Remember")
+
+    # no password, returns encrypted data frame
+    d2 <- data_read(tmp, verbose = FALSE)
+    expect_named(d2, "encrypted")
```
etiennebacher: Check that `d` and `d2` are different

NEWS.md
```diff
@@ -9,6 +9,10 @@ CHANGES
 
 * `standardize()` now works on `fixest` estimations (#665).
 
+* `data_read()` and `data_write()` gain a `password` argument, to encrypt and
+  decrypt data files. This currently only works for R file formats (`.rda`,
+  `.rds`, and `.rdata`).
```
etiennebacher: I think this should mention the method and the openssl function used, and the PR number

etiennebacher (changes requested):
Ok, that sounds reasonable to me. I suggested several changes to docs and tests below.

R/data_read.R
```diff
@@ -25,6 +25,10 @@
 #' might be useful for these formats because labelled numeric variables are then
 #' converted into factors and exported as character columns - else, value labels
 #' would be lost and only numeric values are written to the file.
+#' @param password Password for data encryption. If not `NULL`, the data will be
+#' encrypted (for `data_write()`) or decrypted (for `data_read()`) using the
+#' provided password. Encryption is currently only supported for R file formats
+#' (`.rds`, `.rda` and `.rdata`).
```
etiennebacher:
```suggestion
#' provided password. Encryption is currently only supported for R file formats
#' (`.rds`, `.rda` and `.rdata`). See the section "Data encryption" below for more
#' information on the encryption method used.
```

R/data_read.R
strengejacke: I first switched to sha512, but that doesn't seem to be supported by the openssl-functions. That's why I switched back to sha256:

``` r
passphrase <- charToRaw("This is super secret")
x <- serialize(iris, NULL)

key <- openssl::sha256(passphrase)
y <- openssl::aes_cbc_encrypt(x, key = key)

key <- openssl::sha512(passphrase)
y <- openssl::aes_cbc_encrypt(x, key = key)
#> Error in `aes_any()`:
#> ! key must be of length 16 (aes-128), 24 (aes-192) or 32 (aes-256)
y <- openssl::aes_ctr_encrypt(x, key = key)
#> Error in `aes_any()`:
#> ! key must be of length 16 (aes-128), 24 (aes-192) or 32 (aes-256)
y <- openssl::aes_gcm_encrypt(x, key = key)
#> Error in `aes_any()`:
#> ! key must be of length 16 (aes-128), 24 (aes-192) or 32 (aes-256)
```

<sup>Created on 2026-04-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

strengejacke:
Will continue working on docs/tests later

strengejacke:
Ok, I think all comments are addressed now.

etiennebacher:
Thanks! Just waiting for CI to be green

etiennebacher (approved):

PR #682, opened 2026-05-07
`data_read()` can directly read  zip-files from a URL

strengejacke:


mattansb:
Very nice!

strengejacke:
Not sure why we haven't added this earlier. I'm always working with files that are already downloaded 😜

strengejacke:
I mean, it was basically a one-liner:
```r
path <- .check_path_url(path, file_type = "zip")
```

etiennebacher (approved):
LGTM, I'm investigating the failing CI that seems to be due to `roxygen2` 8.0.0

PR #684, opened 2026-05-20
Add tests from book examples of Discovering statistics

strengejacke:
Example from Chapter 8.2, to ensure we don't introduce any breaking.


etiennebacher:
Can you tweak `easystats/workflows` to add the version constraint on `discovr`?

strengejacke:
I changed the workflow files, but errors are still there. Not sure, maybe I did the wrong modifications?

tests/testthat/test-discovr.R
```diff
@@ -0,0 +1,14 @@
+skip_on_cran()
```
etiennebacher: Why do we want to skip this on CRAN?

etiennebacher:
> I changed the workflow files, but errors are still there. Not sure, maybe I did the wrong modifications?

`discovr` was duplicated there, I fixed it, it should pass CI now

tests/testthat/test-discovr.R
strengejacke: No particular reason, we can also run them on CRAN (though snapshots aren't run on CRAN anyway)

etiennebacher: I removed it just in case we add non-snapshot tests in the future

etiennebacher (approved):

PR #689, opened 2026-06-12
Add `metrics` parameter to `data_tabulate()`

### LINKED ISSUE #685: Cumulative distribution should be optional
Thanks so much for making a useful frequency and cross tabs system! 

Right now on the frequency tables the cumulative distribution is always included. However, in many/most 
cases--meaning when variables are not ordinal--cumulative distribution is both irrelevant and confusing. I'm getting ready to teach this next week and I had to write in the documentation "ignore those columns for now." 

It looks like it would be pretty easy to add a parameter to make the cumulative conditional or to only include those columns when `is.ordered()` or `is.numeric` return `TRUE`. 

https://github.com/easystats/datawizard/blob/main/R/data_tabulate.R#L262

I'd be happy to take a swing at implementing if it is of interest.

elinw:
There were a few ways to approach this. The cumulative values depends on having the valid values, and it could make sense to just set any not included measures to NULL at the end, the way I did for valid.

I added a test for `measures = "raw"` ... I could add for other variations also.

Fixes #685


elinw:
#685

R/data_tabulate.R
```diff
@@ -154,6 +158,7 @@ data_tabulate.default <- function(
   proportions = NULL,
   name = NULL,
   verbose = TRUE,
+  measures = c("raw", "valid", "cumulative"),
```
strengejacke: Can you validate the input, e.g. using `match.arg(..., several.ok = TRUE)`?

R/data_tabulate.R
```diff
@@ -250,7 +255,9 @@ data_tabulate.default <- function(
     out$N <- round(out$N)
   }
 
-  out$`Raw %` <- 100 * out$N / sum(out$N)
+  if ("raw" %in% measures){
```
strengejacke: Please use space around infix-operators :-)

tests/testthat/test-data_tabulate.R
```diff
@@ -181,6 +181,73 @@ test_that("data_tabulate data.frame", {
...
+        unique(efc$e16sex)
+      ),
+      NA
+    ))
+  )
+  expect_identical(table1$N, as.vector(table(addNA(efc$e16sex))))
+  expect_identical(
+    table1$`Raw %`,
+    as.vector(c(
+      100 * table(efc$e16sex, useNA = "always") / length(efc$e16sex)
```
strengejacke, line 242: See lintr, you may remove `c()` here.

R/data_tabulate.R
```diff
@@ -39,6 +39,10 @@
 #' printed as markdown or HTML table, depending on the environment. See
 #' [`insight::export_table()`] for details.
 #' @param verbose Toggle warnings and messages.
+#' @param measures Optional character vector, indicating the types of
+#' percents to be included. Only applies to frequencies, i.e. when `by` is
+#' `NULL`. Can be `"raw"` (includes `NA` values), `"valid"` (excludes `NA` values)
+#' or `"cumulative"` (excludes `NA` vlues).
```
strengejacke: Typo (`vlues`)

etiennebacher (changes requested):
On top of the other comments to address above, please add a bullet point in `NEWS.md` and add an example for the new argument in the docs of `data_tabulate()`.

(@strengejacke I hid all the copilot comments because they were just duplicates of your comments and adding noise to this PR)

R/data_tabulate.R
elinw: Oh missed this, will do this now.

elinw: Is it okay that it returns an error if none of the values given are valid?

elinw:
I was wondering if it should be documented that `measures = c()` will return the frequencies only.
Also, that made me wonder if "N" should also be optional.

R/data_tabulate.R
elinw: This is fixed, not sure why it doesn't show as outdated.

R/data_tabulate.R
```diff
@@ -86,6 +90,9 @@
 #' # drop missing values
 #' data_tabulate(efc$c172code, remove_na = TRUE)
 #'
+#' # exclude the cumulative percent column
+#' data_tabulate(efc$c172code, remove_na = TRUE)
+#'
```
etiennebacher, line 96: The new example should show how to use `measures` instead of `remove_na`.

R/data_tabulate.R
```diff
@@ -39,6 +39,10 @@
 #' printed as markdown or HTML table, depending on the environment. See
 #' [`insight::export_table()`] for details.
 #' @param verbose Toggle warnings and messages.
+#' @param measures Optional character vector, indicating the types of
+#' percents to be included. Only applies to frequencies, i.e. when `by` is
+#' `NULL`. Can be `"raw"` (includes `NA` values), `"valid"` (excludes `NA` values)
+#' or `"cumulative"` (excludes `NA` values).
```
etiennebacher:
```suggestion
#' `NULL`. Can be several choices among `"raw"` (includes `NA` values), 
#' `"valid"` (excludes `NA` values) and `"cumulative"` (excludes `NA` values).
```

tests/testthat/test-data_tabulate.R
```diff
@@ -181,6 +181,73 @@ test_that("data_tabulate data.frame", {
   )
 })
 
+test_that("data_tabulate data.frame with measures", {
```
etiennebacher: Please add tests for `measures = c()`, `measures = NULL` and `measures = "foo"`.

NEWS.md
```diff
@@ -20,6 +20,9 @@ CHANGES
   `.rds`, and `.rdata`). Data encryption is based on the AES-GCM algorithm using
   the `openssl::aes_gcm_encrypt()` function (#675).
 
+* `data_tabulate()` gain a `measures` argument to allow selection of columns to
+  display ("raw", "valid", and "cumulative") (#689).
```
etiennebacher: You can add your github username if you want: `(#689, <username>)`

etiennebacher (commented):
> I was wondering if it should be documented that measures = c() will return the frequencies only.

Yes I think so.

>  Also, that made me wonder if "N" should also be optional.

I don't know about that but if it's not included then I think this `measures` parameter should be renamed `shares`. @strengejacke WDYT?

---

The workflow for code styling fails. We use [Air](https://posit-dev.github.io/air/) to automatically format code so you can install it and run it on the project to fix this failure.

elinw:
> > I was wondering if it should be documented that measures = c() will return the frequencies only.
> 
> Yes I think so.
> 
> > Also, that made me wonder if "N" should also be optional.
> 
> I don't know about that but if it's not included then I think this `measures` parameter should be renamed `shares`. @strengejacke WDYT?
> 
> The workflow for code styling fails. We use [Air](https://posit-dev.github.io/air/) to automatically format code so you can install it and run it on the project to fix this failure.

"shares" is not meaningful to me.  Maybe "percentages" would be more descriptive?

tests/testthat/test-data_tabulate.R
elinw: What do you think should happen for measures = NULL?  In the current design it acts like c(), which makes sense to me. However, just like with measures = "foo" it ignores without erroring.

Tests are now added on that basis.

R/data_tabulate.R
elinw: I changed the wording but not exactly to yours. See what you think.

strengejacke:
If `N` is included, `percentages` would not fit perfectly. I would say `metrics` (if `N` should be included) or `percentages` is fine.

elinw:
> If `N` is included, `percentages` would not fit perfectly. I would say `metrics` (if `N` should be included) or `percentages` is fine.

This sounds good.  Can you let me know if `N` definitely should be included? 
I'll do a commit for the name change alone first.

elinw:
> > If `N` is included, `percentages` would not fit perfectly. I would say `metrics` (if `N` should be included) or `percentages` is fine.
> 
> This sounds good. Can you let me know if `N` definitely should be included? I'll do a commit for the name change alone first.

I added the "N" (in uppercase) option. I did have to change one additional function that assumed that the N column would be present.  All of the tests still pass.

On the tests, the linter does not like testing for `metrics = c()`.  I do think people will write that exactly so it's good to check, but `identical(c(), NULL)` is `TRUE` so technically it's a redundant test.  Let me know what you think.

elinw:
I think everything is resolved unless you have new requests.

R/data_tabulate.R
```diff
@@ -39,6 +39,12 @@
 #' printed as markdown or HTML table, depending on the environment. See
 #' [`insight::export_table()`] for details.
 #' @param verbose Toggle warnings and messages.
+#' @param metrics Optional character vector, indicating the types of
+#' percents to be included. Only applies to frequencies, i.e. when `by` is
+#' `NULL`. Can include any combination of `N` (frequencies including NA), `"raw"` (includes `NA` values),
+#' `"valid"` (excludes `NA` values) or `"cumulative"` (excludes `NA` values).
+#' Using `c()` will return a table with only the value labels. Invalid
+#' values (`metrics = "foo"`) are silently ignored.
```
etiennebacher:
```suggestion
#' @param metrics Optional character vector, indicating the types of
#' metrics to be included. Only applies to frequencies, i.e. when `by` is
#' `NULL`. Can include any combination of `N` (frequencies including `NA`), 
#' `"raw"` (percentage including `NA` values), `"valid"` (percentage excluding
#' `NA` values) or `"cumulative"` (percentage excluding `NA` values).
```
The part about silently ignoring unknown values seemed wrong (it errors) so I removed it.

R/data_tabulate.R
```diff
@@ -86,6 +92,12 @@
 #' # drop missing values
 #' data_tabulate(efc$c172code, remove_na = TRUE)
 #'
+#' # exclude the cumulative percent column
+#' data_tabulate(efc$c172code, metrics = c("raw", "valid"))
+#'
+#' # return frequencies only
+#' data_tabulate(efc$c172code, remove_na = TRUE, metrics = c())
```
etiennebacher: The comment is outdated but I don't think we should showcase the `metrics = c()` anyway, it's just an edge case. I think these two lines can be removed.

R/data_tabulate.R
```diff
@@ -276,6 +304,10 @@ data_tabulate.default <- function(
     }
     out <- cbind(var_info, out)
   }
+  total_n <- sum(out$N, na.rm = TRUE)
```
etiennebacher, line 317: Why has this line moved? I don't think it causes an issue in itself but I'd rather minimize the git diff.

NEWS.md
```diff
@@ -20,6 +20,9 @@ CHANGES
   `.rds`, and `.rdata`). Data encryption is based on the AES-GCM algorithm using
   the `openssl::aes_gcm_encrypt()` function (#675).
 
+* `data_tabulate()` gain a `measures` argument to allow selection of columns to
+  display ("raw", "valid", and "cumulative") (#689, @elinw).
```
etiennebacher:
```suggestion
* `data_tabulate()` gain a `metrics` argument to allow selection of columns to
  display (#689, @elinw).
```

etiennebacher (changes requested):

etiennebacher:
Can you also resolve the conflicts?

elinw:
I tried running Air on the file but the only issue it identified was duplicate blank lines. Not sure why this is failing.

etiennebacher:
The problem was that a new version of Air was released a couple of days ago and replaces `=` by `<-` by default (when possible). This was clashing with our custom syntax that can be used in `data_filter()`.

There are just the few comments above to resolve now (don't worry about linting failures).

elinw:
> The problem was that a new version of Air was released a couple of days ago and replaces `=` by `<-` by default (when possible). This was clashing with our custom syntax that can be used in `data_filter()`.
> 
> There are just the few comments above to resolve now (don't worry about linting failures).

How about having a separate branch for those, since the comments are in unrelated files?

etiennebacher:
Yes I have actually merged a PR that fixes those Air failures so that these changes don't appear here. Your latest commit message says you left questions on some comments but I don't see anything, am I missing something? I just want to avoid a situation where we're both waiting for something from the other.

elinw:
> Yes I have actually merged a PR that fixes those Air failures so that these changes don't appear here. Your latest commit message says you left questions on some comments but I don't see anything, am I missing something? I just want to avoid a situation where we're both waiting for something from the other.

I think, in the end, the only question I had was about metrics = "foo" and that doesn't matter.

etiennebacher:
There are still unresolved comments above about docs/NEWS/line changed. I have made "suggestions" using Github interface but didn't actually apply them.

elinw:
I'm not sure why two builds are failing, but they don't seem related. R crashed during the test of ` test-standardize_models.R: ` but it might be that retriggering the builds would fix. Not sure what would have changed. 
R 4.6.1 was released today so I guess it couldbe related to that.

Update:

── Warning (test-standardize-data.R:23:1): standardize.numeric ─────────────────
Global state has changed:
     names(before[[4]]) | names(after[[4]])      
[79] "TESTTHAT"         | "TESTTHAT"         [79]
[80] "TESTTHAT_WD"      | "TESTTHAT_WD"      [80]
[81] "TMPDIR"           | "TMPDIR"           [81]
                        - "TZDIR"            [82]
[82] "USER"             | "USER"             [83]
[83] "XPC_FLAGS"        | "XPC_FLAGS"        [84]
[84] "XPC_SERVICE_NAME" | "XPC_SERVICE_NAME" [85]
before[[4]][79:84] vs after[[4]][79:85]
  TERM"xterm-256color"
  TESTTHAT"true"
  TESTTHAT_WD"/Users/elinwaring/Code/rcode/datawizard"
  TMPDIR"/var/folders/gw/_qnrkpdn3fjg_0gwwnb9h_g40000gn/T/"
+ TZDIR"/var/db/timezone/zoneinfo"
  USER"elinwaring"
  XPC_FLAGS"0x0"
  XPC_SERVICE_NAME"application.com.rstudio.desktop.371427447.371427452"                                                                                                                                                             
[ FAIL 0 | WARN 1 | SKIP 0 | PASS 75 ]

update 2

https://discourse.mc-stan.org/t/global-state-has-changed-warning/36591

Was this resolved for `report`?  

update 3

Apparently the resolved this by deleting `tests/testthat/helper-state.R`

https://github.com/easystats/report/pull/471

etiennebacher:
Failure on R 4.1 comes from https://github.com/mitchelloharawild/distributional/issues/157

etiennebacher (approved):
Thanks!

PR #690, opened 2026-06-12
Add a `by` attribute to `data_tabulate()` objects when `by` is present

### LINKED ISSUE #688: Add `by` attribute to cross_tab and cross_tabs objects
Right now when you get a cross_tab or cross_tabs object the name of the "by" variable is not stored as an attribute. This means   that down the line you don't have it available for labels and post processing (and is potentially confusing when many variables have the same value names).

Would you consider adding a by attributes?

### Example crosstabs
```
 data_tabulate(efc, c("e42dep"), by = "c161sex") ->a
 attributes(a)
attributes(a[[1]])
```
```
$class
[1] "datawizard_crosstabs" "list"                

$collapse
[1] FALSE

$is_weighted
[1] FALSE
```
```
attributes(a[[1]])
```
```
$names
[1] "e42dep" "Male"   "Female" "NA"    

$row.names
[1] 1 2 3 4 5

$class
[1] "datawizard_crosstab" "data.frame"         

$total_n
[1] 908

$varname
[1] "e42dep"

$grouped_df
[1] FALSE
```
### Example crosstab
```
data_tabulate(efc$e42dep, by = efc$c161sex) ->b
attributes(b)
```
```
$names
[1] "efc$e42dep" "Male"       "Female"     "NA"        

$row.names
[1] 1 2 3 4 5

$class
[1] "datawizard_crosstab" "data.frame"         

$total_n
[1] 908

$varname
[1] "efc$e42dep"

$grouped_df
[1] FALSE

```

elinw:
I wasn't sure if you would want an additional test; all the current tests are still passing.
The attributes aren't listing in the documentation so I didn't change anything there.

Closes #688

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ CHANGES
 
 * `data_read()` now also reads zip-files from URLs (#682).
 
+* `data_tabluate()` now returns an attribute "by" with the
+  `by` variable name when the `by` parameter is used.
+
 # datawizard 1.3.1
 
 CHANGES
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,8 +4,8 @@ CHANGES
 
 * `data_read()` now also reads zip-files from URLs (#682).
 
-* `data_tabluate()` now returns an attribute "by" with the
-  `by` variable name when the `by` parameter is used.
+* `data_tabulate()` now returns an attribute "by" with the
+  `by` variable name when the `by` parameter is used (#690 @elinw).
 
 # datawizard 1.3.1
 
--- a/R/data_tabulate.R
+++ b/R/data_tabulate.R
@@ -39,6 +39,10 @@
 #' printed as markdown or HTML table, depending on the environment. See
 #' [`insight::export_table()`] for details.
 #' @param verbose Toggle warnings and messages.
+#' @param measures Optional character vector, indicating the types of
+#' percents to be included. Only applies to frequencies, i.e. when `by` is
+#' `NULL`. Can be `"raw"` (includes `NA` values), `"valid"` (excludes `NA` values)
+#' or `"cumulative"` (excludes `NA` vlues).
 #' @param ... not used.
 #' @inheritParams extract_column_names
 #'
@@ -154,6 +158,7 @@ data_tabulate.default <- function(
   proportions = NULL,
   name = NULL,
   verbose = TRUE,
+  measures = c("raw", "valid", "cumulative"),
   ...
 ) {
   # save label attribute, before it gets lost...
@@ -250,7 +255,9 @@ data_tabulate.default <- function(
     out$N <- round(out$N)
   }
 
-  out$`Raw %` <- 100 * out$N / sum(out$N)
+  if ("raw" %in% measures){
+    out$`Raw %` <- 100 * out$N / sum(out$N)
+  }
   # if we have missing values, we add a row with NA
   if (remove_na) {
     out$`Valid %` <- 100 * out$N / sum(out$N)
@@ -259,8 +266,14 @@ data_tabulate.default <- function(
     out$`Valid %` <- c(100 * out$N[-nrow(out)] / sum(out$N[-nrow(out)]), NA)
     valid_n <- sum(out$N[-length(out$N)], na.rm = TRUE)
   }
-  out$`Cumulative %` <- cumsum(out$`Valid %`)
 
+  if ("cumulative" %in% measures) {
+    out$`Cumulative %` <- cumsum(out$`Valid %`)
+  }
+
+  if (!"valid" %in% measures) {
+     out$`Valid %` <- NULL
+  }
   # add information about variable/group names
   if (!is.null(obj_name)) {
     if (is.null(group_variable)) {
--- a/R/data_tabulate.R
+++ b/R/data_tabulate.R
@@ -203,8 +203,7 @@ data_tabulate.default <- function(
       proportions = proportions,
       obj_name = obj_name,
       group_variable = group_variable
-    )
-    )
+    ))
   }
   # frequency table
   if (is.null(weights)) {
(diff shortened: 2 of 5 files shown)


R/data_tabulate.R
```diff
@@ -345,6 +354,7 @@ data_tabulate.data.frame <- function(
     class(out) <- c("datawizard_tables", "list")
   } else {
     class(out) <- c("datawizard_crosstabs", "list")
+    attr(out, "by") <- gsub('\\"', "", by_name, fixed = TRUE)
```
etiennebacher: Why is the `gsub()` necessary?

R/data_tabulate.R
```diff
@@ -416,6 +431,7 @@ data_tabulate.grouped_df <- function(
...
```
etiennebacher: Same

NEWS.md
```diff
@@ -4,6 +4,9 @@ CHANGES
 
 * `data_read()` now also reads zip-files from URLs (#682).
 
+* `data_tabluate()` now returns an attribute "by" with the
+  `by` variable name when the `by` parameter is used.
```
etiennebacher:
```suggestion
* `data_tabulate()` now returns an attribute "by" with the
  `by` variable name when the `by` parameter is used.
```

Also please add the PR number and your username.

etiennebacher (changes requested):
Thanks! Yes please add at least one test for this.

R/data_tabulate.R
elinw: Because in this case (but not the other) the name returns e.g. "/"Island"/".  I'm not sure why or if there is a way to get the name without the escapes. I tried some ideas but there was nothing obvious to me.

R/data_tabulate.R
elinw: See above

R/data_tabulate.R
elinw: Meaning when a quoted variable name is used for by.

NEWS.md
elinw: Where do you want the username? As @elinw or elinw?

etiennebacher: @elinw

R/data_tabulate.R
etiennebacher: Ok, should this also remove `'` if one passes `'Island'` in `by`?

etiennebacher: I also don't think `\\` is necessary if `fixed = TRUE` is used

elinw: And actually I stupidly followed the advice from the tester and added fixed = TRUE but that was wrong, this is not fixed it just looks like fixed. So the next push will put that back.

elinw: Yeah it's not a bad instinct but it needs the escaping to work for whatever reason. But in the update I'm about to send you will see a change.
I just tried it again and it makes the tests fail to use fixed=TRUE

elinw:
Also the linter wanted me to make some changes and then not make them (when I made them it told me to change them back) unless I misunderstood.

R/data_tabulate.R
etiennebacher: Sorry but I don't see a case where using the backslash changes something:

``` r
x <- insight::safe_deparse("foo")
x
#> [1] "\"foo\""

gsub('"', "", x)
#> [1] "foo"
gsub('"', "", x, fixed = TRUE)
#> [1] "foo"
gsub('\\"', "", x)
#> [1] "foo"
```

Do you have an example where this affects the output?

R/data_tabulate.R
```diff
@@ -415,6 +450,7 @@ data_tabulate.grouped_df <- function(
   if (is.null(by)) {
     class(out) <- c("datawizard_tables", "list")
   } else {
+    out <- lapply(out, structure, by = by_name)
```
etiennebacher, line 452: I don't understand why this attribute needs to be added to each element in `out`. Below we add the attribute `collapse` to `out` only, why can't we do that for `by` since it has the same value for all elements in `out` anyway?

tests/testthat/test-data_tabulate.R
```diff
@@ -181,7 +181,78 @@ test_that("data_tabulate data.frame", {
   )
 })
 
+test_that("data_tabulate data.frame by", {
+  data(efc, package = "datawizard")
+  x <- data_tabulate.data.frame(efc, "c172code", by = "e16sex")
```
etiennebacher:
```suggestion
  x <- data_tabulate(efc, "c172code", by = "e16sex")
```

`data_tabulate()` is an S3 generic so it should be dispatched automatically to `data_tabulate.data.frame()` when the input is a `data.frame`.

etiennebacher (changes requested):

R/data_tabulate.R
elinw: Yes this is what took me a long time to figure out.  "by"  is really a property of individual tables not of the list.  This should be consistent whether you get the table through the data.frame method or through the default method or if select has one or multiple columns. Also it should be the same for grouped data frames.

It took me a while to see this, which was why I would get the tests to work on default and then they would fail or data.frame or with multiple columns.
As a practical matter, if you pull out one table from the list you should get the "by" along with it.

R/data_tabulate.R
elinw: Yes fixed.

elinw:
I don't love the way this renders but the links  do work

elinw:
wrong branch

elinw:
I'm not sure if there is a change I missed but it is still showing one requested change . I can't tell what it is or if it was actually resolved.

R/data_tabulate.R
etiennebacher: Ok I'm convinced.

(Re your comment about "Requested changes" below, this is because Github keeps this label until the same person approves the PR.)

etiennebacher (approved):

strengejacke:
Please remember to increase the version number in the fourth place by one for "user visible" changes (e.g. to 1.3.1.x) in the description file. This ensures that `easystats::install_latest()` Will work as intended.

PR #691, opened 2026-06-16
Fix parameters for contr.deviation() 

elinw:
Currently `contr.deviation() ` is using `@inheritParams` from `stats::contr.sum()`, but because of how those parameters are written using the new style for linking S4 classes this is raising errors. I raised this as an issue in `stats`, but they said that this should be addressed in the package.


elinw:
messed up branches

PR #692, opened 2026-06-16
Fix params in `contr.deviation()`

elinw:
Currently contr.deviation()  is using @inheritParams from stats::contr.sum(), but because of how those parameters are written using the new style for linking S4 classes this is raising errors. I raised this as an issue in stats, but they said that this should be addressed in the package.

### DIFF
--- a/R/contrs.R
+++ b/R/contrs.R
@@ -7,8 +7,8 @@
 #'  baseline group. Ignored if `contrasts` is `FALSE`.
 #' @param contrasts description
 #' @param sparse logical indicating if the result should be sparse
-#' (of class [`Matrix`](linkS4class(dgCMatrix)), using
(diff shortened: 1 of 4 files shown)


etiennebacher (changes requested):
Thanks, I would have liked https://github.com/r-lib/roxygen2/issues/1877 to be fixed but I don't know when the next `roxygen2` update will be. 

Can you remove unrelated changes in `data_tabulate.R` and `data_tabulate.Rd`?

elinw:
I thought i got rid of them.
Basically R core says if you want to use that function you have to to an R 4.6 minimum which is extreme in my opinion.

elinw:
Okay I had to do so some git magic but there is now a clean version at #693 693

PR #693, opened 2026-06-17
Stop using `@inheritsParams stats::contr.sum`

elinw:


R/contrs.R
```diff
@@ -2,7 +2,13 @@
 #'
 #' Build a deviation contrast matrix, a type of _effects contrast_ matrix.
 #'
-#' @inheritParams stats::contr.sum
+#' @param n a vector of levels for a factor, or the number of levels.
+#' @param base an integer specifying which group is considered the
+#'  baseline group. Ignored if `contrasts` is `FALSE`.
+#' @param contrasts description
```
etiennebacher: Missing the actual description?

etiennebacher (changes requested):

etiennebacher:
Also please use a capital letter for each start of param description

R/contrs.R
elinw: Are you sure that `base` is supposed to be here? 
I looked at https://search.r-project.org/R/refmans/stats/html/contrast.html  and for contr.sum() it only shows 3.

etiennebacher: I don't understand what you mean, I just checked that the generated `.Rd` file only has minor changes and that CI doesn't fail anymore.

etiennebacher (approved):
Thank you!

PR #699, opened 2026-07-06
Assign dimnames in as.table.datawizard_crosstab()

elinw:
Currently the` as.table.datawizard_crosstab()` function produces a table in which the `dimnames` are unnamed.  When making a table using `table()` the names are created, but as.table() does not have an argument to do it directly. Having the dimnames set is very helpful when using tables for further analysis (like `loglin()`).


etiennebacher (changes requested):
Thanks, LGTM, can you add an item in `NEWS.md` and bump the last version number in `DESCRIPTION`?

elinw:
I wasn't sure what level of version change you wanted.

etiennebacher (approved):

PR #700, opened 2026-07-17
Updates to fix build warnings.

elinw:


etiennebacher:
I see a `NOTE` related to the docs in previous R CMD checks but I don't see why the changes in test files are needed, can you explain why you made those?

strengejacke:
We used `dput()` to create the factor output, which returned a `structure()`. This code was created some time ago, using a now deprecated name. If you run `dput()` on a factor now, it returns an element `levels`, not `.Label`. so it's just about replacing deprecated names.

strengejacke:
Note, however, that `labels` and `levels` are not always interchangeable, because labels gives you new labels, even if these did not exist as values before (and assigns those as correct levels), while levels only works when the levels are also present as values.

etiennebacher (approved):
Thanks!

>  If you run dput() on a factor now, it returns an element levels, not .Label. so it's just about replacing deprecated names.

Interesting, but did this come up in some checks? I'm just wondering how you ended up noticing this was needed.

strengejacke:
I visit https://easystats.github.io/dashboard/dashboard.html every now and then

PR #701, opened 2026-07-26
Add function for chi square.

elinw:
This function follows the pattern in the documentation, and I found it useful to create it for my students.  It might be good to let it make a caption the way that the `means_by_groups()` has the ANOVA results.

### DIFF
--- a/R/data_chisq.R
+++ b/R/data_chisq.R
@@ -1,0 +1,15 @@
+#' Calculate chi square test for tabulate
+#'
+#' @param dwtable    A dwtable object.
+#' @param show_table Should the table be shown with the chi sqare
+#' @param ...  Other options
+#'
+#' @export
+data_chisq <- function(dwtable, show_table = TRUE, ...) {
+  tablelist <- as.table(dwtable)
+  chisqlist <- lapply(tablelist, stats::chisq.test)
+  if (length(chisqlist) == 1) {
+    return(chisqlist[[1]])
+  }
+  chisqlist
+}


strengejacke:
You may take a look at the _effectsize_ package where you can calculate lots of effect sizes from different R objects.

strengejacke:
Ok, I see the intention of this function. Maybe it's better to implement a `chisq.test.dwtable()` method instead? @etiennebacher wdyt?

etiennebacher:
Personally I don't really see the need for this function, it seems to simply replace a call to `as.table()` and `lapply()`? I get that it makes it easier for students but we also can't implement a function for all possible computations one could do on a table.

``` r
library(datawizard)

### Without data_chisq()
suppressWarnings({
  orig <- head(mtcars) |> 
    data_tabulate() |> 
    as.table() |> 
    lapply(stats::chisq.test)
})

### With data_chisq()
suppressWarnings({
  new <- head(mtcars) |> 
    data_tabulate() |> 
    data_chisq()
})

identical(orig, new)
#> [1] TRUE
```


> Maybe it's better to implement a chisq.test.dwtable() method instead? 

I would be in favor of adding this as a method, but `chisq.test()` isn't generic:

``` r
sloop::is_s3_generic("chisq.test")
#> [1] FALSE
```

elinw:
Sorry for the delay in responding, I was away on vacation for a few weeks, will think about the comments.  
My main idea, though, is have the caption (that is be able to print the table with the chi square).  I was thinking that it would be like some of the other data_* methods.

elinw:
I came across https://datacolada.org/132  the statuser which has a table2 function that has an option for requesting chi square. Maybe the thing to do is to add the Chi Square as an attribute to the data_tabulate object -- the cost of the performance would probably be worth it because of making it simpler for the user and for subsequent processing.

PR #702, opened 2026-08-29
distribution_mode in somewhat complex cases

### LINKED ISSUE #695: distribution_mode in somewhat complex cases
If you try some special cases of the mode, I think there are incorrect results.

```
> distribution_mode(c(1, 2, 3, 3, 4, 4, 5))
[1] 3
> distribution_mode(c(1, 2, 3, 4, 5))
[1] 1
> distribution_mode(c(1, 2, 3, 3, 4, 4, 5, 5))
[1] 3
> distribution_mode(c(1, 2, 3, 3, 4, 5, 5))
[1] 3

```

On a simple level the issue is what should happen if there are ties or if all of the frequencies are tied.  The simplest thing would be to return `NA` in those cases. 

This long Stack Overflow thread
http://stackoverflow.com/questions/2547402/standard-library-function-in-r-for-finding-the-mode

discusses some of the issues.

Update
Here is a link to what SASdoes.
https://documentation.sas.com/doc/en/pgmsascdc/v_073/procstat/procstat_univariate_examples02.htm
Basically, like the current function, when there are multiple modes in a set of statistics they return the lowest but including a note that indicates how many modes there are. 
When there are multiple modes and the mode alone is selected they return a table with all of them.  (When I made a mode function previously I returned a vector of modes.)

strengejacke:
Fixes #695

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: datawizard
 Title: Easy Data Wrangling and Statistical Transformations
-Version: 1.3.1.3
+Version: 1.3.1.4
 Authors@R: c(
     person("Indrajeet", "Patil", , "patilindrajeet.science@gmail.com", role = "aut",
            comment = c(ORCID = "0000-0003-1995-6531")),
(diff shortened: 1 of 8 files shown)


etiennebacher (approved):
Thanks, LGTM

strengejacke:
Thanks! WDYT about a new CRAN release?

etiennebacher:
I can release 1.4.0 this week if you want, I don't think this requires other releases to be done first.

strengejacke:
Great! No, it doesn't depend on other releases, so there's no specific timing / schedule for submission

